PR #2, opened 2023-06-23
Mod: Language filter

artillect:
This is an updated version of my [greasyfork script](https://greasyfork.org/en/scripts/468754-kbin-language-filter) that filters posts based on the language indicated in their badge. I might need to make some adjustments to account for OC posts, and untagged posts (in case people want to filter out all English posts)


aclist:
Thank you, will test and revert back. Question for you: the initial design calls for the remote scripts (mods) to be archived at this repository under the `/mods` dir so that the megamod calling script fetches the resources from a single endpoint.

Rationale: mods are fetched from the remote resource URL and not actually "installed" by the user. They are loaded at runtime the first time and then cached locally.  While, e.g., vim plugin managers like Plug and Vundle let users specify arbitrary remote repos, I feel like the target user is more technically-minded and they can audit their own plugins list.

With this, users are loading the single megamod script and might not see/understand where the remote files are coming from. I feel this represents a (very small) security risk because the author of the upstream script could make changes that the megamod user isn't aware of. Granted, that's true when installing any script off of greasyfork, too, but I feel like we should apply the principle of least trust here and whitelist the incoming mods by archiving them into the `/mods` dir as a copy so that they are immune to any change at the remote destination. 

This way we can be sure that the mods being served through megamod have been audited and have not undergone any random changes. What do you think?

Regarding GUI creation, I have started a new discussion thread here: https://github.com/aclist/kbin-megamod/discussions/3

aclist:
As a follow-up to the above, since it's reasonable to assume that every mod author isn't going to bother porting their scripts, I think it also makes sense to archive the good ones into megamod (with proper attribution, of course) and extend them by adding toggle functionality.

artillect:
I think it'd be a great idea to have all of the mods contained in this repository under the `/mods` dir, makes sense security-wise, and it should hopefully be pretty easy to encourage script authors to port their scripts into here. Speaking of attribution, I feel like you could add a field for each mod's author so they show up in the settings next to the toggles.

megamod.user.js
```diff
@@ -15,6 +15,7 @@
 // @require       https://github.com/aclist/kbin-megamod/raw/main/mods/label.user.js
 // @require       https://github.com/aclist/kbin-megamod/raw/main/mods/dropdown.user.js
 // @require       https://github.com/aclist/kbin-megamod/raw/main/mods/code-highlighting.user.js
+// @require       https://github.com/artillect/kbin-megamod/raw/main/mods/language-filter.user.js
```
aclist, line 18: Can you add this file verbatim to the source repo and update this line?

aclist (changes requested):
Request to add script to mods dir and update @require

artillect:
I think I sorted out the @require, but what do you mean by adding the file to the source repo? Shouldn't it get added once you merge the pull request?

PS: sorry about all the extra commits that just got added, I'm still working on learning git. I should've made a separate branch in my fork for this, I'll make sure to do that next time

aclist:
You are correct, sorry, I blanked out while doing the review and didn't notice the adjoining file for some reason. All you had to do was update that includes line! 

Yes, agree, best to make a patch branch instead of pushing into main so that the maintainer can pull and test the branch before approval.

aclist:
On initial testing, this works when newly enabling the filter to ON, but if the filter is already to ON and the page is reloaded, it is not triggering as intended. Possible race condition? I'll try debugging it later. Basically, the code should be stateful so that previously ticked options are sticky and persist sessions.

Off the top of my head, possible fixes might be: use the method @run-at document-end or load waitForKeyElements.js

artillect:
I think there's an issue with the dropdown mod trying to disable itself that causes the rest of the mods to stop loading. Not sure how to fix that, but I just commented out the lines enabling it in megamod.user.js to get everything else after it to work

aclist:
OK, great, duly noted, that is on me. The dropdown is buggy and I need to look at it. We should probably use a try-catch statement to allow other mods to pass if one of them fails.

PR #7, opened 2023-06-26
Add version check

artillect:


aclist:
Just came across this method that should be straightforward for ingesting text in flat files: https://sourceforge.net/p/greasemonkey/wiki/GM_getResourceText/

artillect:
For some reason, making this change also made it so that the version check actually happens properly the first time you open the modal. I think it's almost good to go now, I'm just gonna add some final touches

artillect:
Here's what it looks like now
image

aclist:
I was thinking that the existence of a version update should hijack the base element itself so that an update overrides the original link. Thoughts?

artillect:
I think it could work either way, I'm not too attached to either method. I think the biggest thing is that it needs to be clear that you can click it to install the update. I think it should get updated automatically anyways, so we might not even need it in the end. (It might be better to have the version URL be a const instead of in the meta block, so it will check for updates each time the userscript runs rather than checking each time it checks for an update, due to the caching.)

aclist:
I don't think we can make it auto update unless the user explicitly enables this in their *monkey settings, but you're right that we can force it to check each time. 

- Version identifier: just looks like plaintext
- Update available: override this space and style it as a colored button

artillect:
It probably depends on the extension, but I think tampermonkey checks daily by default (unless I've changed it, but I haven't touched my settings on this browser). 

> - Version identifier: just looks like plaintext
> - Update available: override this space and style it as a colored button

Sounds like a plan, I'll change that now

aclist:
I just went ahead and merged it since it's non-destructive, but this still needs to handle race conditions on load:

- Load the page
- New version is caught but link does not get updated when opening the modal
- Close the modal and reopen
- Link updates correctly

Also looks like the default button style in kbin isn't really a button, just a color? This needs to more explicitly look like a button IMO.

PR #8, opened 2023-06-26
WIP: Add pagination to modal UI and adjust layout

artillect:
image


aclist:
Managed to make time to clean up my changes and consolidate. Can you enable this feature so I can bump my changes back to your patch fork? 

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

aclist:
That looks nice. I think it'd put the author at the top, as the description text may look like a short and manageable title for now, but it's actually liable to grow as the helpbox gets tutorialized and includes detailed description of how a certain mod is to be used. E.g.,

I'm thinking that helpbox returns an object consisting of all the types and values set within it and that gets processed by the receiving function accordingly. If toggle key in the function is set to off, then the mod just proceeds to the disable logic. Otherwise, it processes the rest of the parameters. This should make it more generic so that you can just pass the contents of the helpbox selections as-is without hardcoding them for a specific mod, and don't have to process the toggle checkbox separately. 

On that note, what if the checkbox itself is a toggle button (ON/OFF) and goes inside of the helpbox? Then the list items are just clickable links? Should be less fiddly than the current method. And the helpbox is permanently visible, since every mod necessarily contains at least a toggle and description, and potentially more granular options. There can be no ambiguity here and the user isn't hunting for the open-helpbox button.

Focus the active sidebar page/active mod by changing the color, and add the mod name to the crumbs:

b

artillect:
I think that looks pretty great, but I think moving the toggle into the helpbox might hide it a bit too much, especially considering that a lot of the mods are just going to have the enable switch and no other options. I think passing all the values as an object should work pretty well. I think this is getting close to being done!

aclist:
Sorry, I took some time off yesterday. Looks like you guys have been busy.

On the toggle, right, but I think that consitently requiring users to open the settings pane to select the toggle ensures that they are forced to read the detailed description whether or not there are more granular options. The detailed description may contain important information such as "login required to use this setting," etc., whereas just having the toggle there reinforces the behavior of just toggling things without knowing what they do and then being potentially frustrated with the functionality. If every mod always has a unified helpbox area as the main point of interaction, the messaging is more consistent imo. We can always change this around later; it is quite minor.

Anyway, yes, I think this is pretty much there and we should freeze the UI code after I apply the next patch shown in the image above. I'm also going to just merge this PR now since it's the most stable version and so that the commit log doesn't balloon out of control for this single patch. So:

- This UI version will be the most current for now
- Will add a new small UI patch probably today with the toggles 
- Following that, will add a patch for the detailed config options and bump the major version since introduces breaking changes to the "API"

Recipient scripts should expect to receive an object containing the toggle state followed by the parameters set in the config:

`{state:"on", field1: "myhexcolor", field2:"foo"};`

To make this more generic, the fields are just enumerated in incremental order based on what was set up in the config menu. So the `manifest.json`can call as many input fields as it wants, then it gets back whatever was put into them.  It's the responsibility of the recipient script to parse this and do set up. If the script only has a toggle state and nothing more, it can stop at the first key. I think this is going to be OK, will post a discussion topic if anything seems problematic.

PR #16, opened 2023-06-28
Mod: Collapsible comments

artillect:
Allows you to collapse comments by clicking on their headers, or the bars to their left

image

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -74,5 +74,20 @@
       "entrypoint": "languageFilterEntry",
         "type": "textarea",
          "page": "threads"
+    },
+    {
+      "name": "Collapsible comments",
+      "author": "artillect",
+      "version": "1.3.7",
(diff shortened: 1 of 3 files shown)


aclist:
Collisions might be an issue going forward. Perhaps classnames should be namespaced to the name of the mod.

artillect:
> Perhaps classnames should be namespaced to the name of the mod.

I think that's a good call. 

This should be ready to go now that I've sorted the collision issue out, it has a few small issues but nothing that seriously interferes with anything as far as I can tell

aclist:
Maybe, although it necessitates a non-trivial amount of refactoring of other scripts. Presumably we have to do integration testing of them anyway, so it's six of one and half a dozen of the other, I guess.

Will take a look at this tomorrow and seek to finish implementing the multi-config options over the weekend.

PR #17, opened 2023-06-29
Easy Emoticon

minnieoo:
Enables you to use various slash commands for emoticons. [Command guide here.](https://docs.google.com/document/d/1jMeY7GwmgwrV8mEC4jmkEKKWdKuWvjgLgYrtUS5hPYg/edit)

Easy Emoticon examples on a text area

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -80,7 +80,7 @@
         "author": "minnieo",
             "version": "0.6.1",
         "label": "Easy emoticons",
-        "desc": "Use slash commands for emoticons",
+        "desc": "Use slash commands for emoticons. <a href="https://docs.google.com/document/d/1jMeY7GwmgwrV8mEC4jmkEKKWdKuWvjgLgYrtUS5hPYg/edit">Command guide here.</a>",
         "class": "easyEmoticon",
         "entrypoint": "easyEmoticon",
           "type": "checkbox",
--- a/manifest.json
+++ b/manifest.json
@@ -74,5 +74,16 @@
       "entrypoint": "languageFilterEntry",
         "type": "textarea",
          "page": "threads"
-    }
+    },
+        {
+      "name": "Kbin Easy Emoticon",
+        "author": "minnieo",
+            "version": "0.6.1",
+        "label": "Easy emoticons",
+        "desc": "Use slash commands for emoticons",
+        "class": "easyEmoticon",
+        "entrypoint": "easyEmoticon",
+          "type": "checkbox",
+          "page": "general"   
+        }
    ]
--- a/mods/easy-emoticon.user.js
+++ b/mods/easy-emoticon.user.js
@@ -8,6 +8,7 @@
 // @license      MIT
 // ==/UserScript==
 
+let eventListener;
 //lemmy based emoticons
 const emoticons = [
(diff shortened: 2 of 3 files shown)


minnieoo:
this is my first ever PR, so i hope i did it right (ㅅ´ ˘ `)

mods/easy-emoticon.user.js
```diff
@@ -0,0 +1,146 @@
...
+        ;
+
+    param.value = text;
+
+}
+
+function easyEmoticon(toggle) {
+    if (toggle) {
+        emoticonGen();
+    }
```
Oricul, line 145: I think you should look at setting a global variable and setting it to your event listener inside emoticonGen(). This would allow you to remove the listener when toggled off.

minnieoo: done, thanks for the advice!

Oricul (approved):
Looks good to me.

manifest.json
```diff
@@ -74,5 +74,16 @@
...
         "type": "textarea",
          "page": "threads"
-    }
+    },
+        {
+      "name": "Kbin Easy Emoticon",
+        "author": "minnieo",
+            "version": "0.6.1",
+        "label": "Easy emoticons",
+        "desc": "Use slash commands for emoticons",
```
artillect, line 83: You can add a link to the guide by adding `<a href="https://docs.google.com/document/d/1jMeY7GwmgwrV8mEC4jmkEKKWdKuWvjgLgYrtUS5hPYg/edit">Command guide here</a>` to the description

mods/easy-emoticon.user.js
```diff
@@ -0,0 +1,150 @@
...
+            emoticonMake(e.target);
+            console.log('working');
+        }
+    };
+    document.addEventListener('input', eventListener)
+}
+
+function emoticonMake(param) {
+    let text = param.value;
+    text = text
```
artillect, line 47: Instead of using all of these `.replace()`s, you can create a dictionary of commands and emoticons, and use a for loop to replace all of the commands with their respective emoticons. I'm not 100% sure if this would work properly but this should be a good start for cleaning this up and making it easier to add more emoticons:

```
// Replace the emoticons array with a dictionary
const emoticons = {
    /\/shrug/ = '¯\\\\_(ツ)\\_/¯',
    /\/lenny/ = '( ͡° ͜ʖ ͡°)',
    /\/lenshrug/ = '¯\\_( ͡° ͜ʖ ͡°)_/¯',
    // etc.
};

function emoticonMake(param) {
    let text = param.value;

    for (const [command, emoticon] of Object.entries(emoticons)) {
        text.replace(command, `${emoticon} `);
    }

    param.value = text;
}

artillect:
Looks good to me! Let me know if you have any questions about my suggestions, but I think they should just work

mods/easy-emoticon.user.js
minnieoo: woooah that is cleaner. on it! thanks for helping a beginner learn new things ✧⁺⸜(●′▾‵●)⸝⁺✧'

aclist:
Looks great, thanks! The external doc is a good way of handling detailed documentation that would be too lengthy otherwise.

PR #18, opened 2023-06-29
kbin-code-highlighting upgrade to 0.4.2 and load/unload functionality.

Oricul:
Numerous edits around KCH. Biggest of which is updated to 0.4.2. Was able to get this working with the current iteration of megamod.


aclist:
This is looking great, a predefined list of styles is something I was missing in the first iteration. I'll look into externalizing this object at the top of the file so that drop-in addition of mods does not require modifying the base script.

PR #24, opened 2023-07-01
fix: Replaced test version with actual version check

artillect:
I accidentally left a little change I made for debugging in my pull request (setting the new version to 1.0 to force the update button to appear), this returns it to normal


aclist:
I noticed that but wasn't being too thorough in the audit process and just wanted to merge it. Sorry to make you update again. You could also downgrade your local version number to trigger a change.

PR #27, opened 2023-07-01
Mod: Hide votes and reputation

artillect:
Adds toggles for hiding favorites (upvotes), reduces (downvotes), and reputation. This should cover one of the features in #14


aclist:
I'm getting some bug where the site isn't letting me approve this PR, but it has in fact been merged already, so closing.

PR #29, opened 2023-07-01
Mod: Show non-local instance names next to users or communities

artillect:
Adds toggles to show instance names next to users and magazines/communities

image


aclist:
The function `hideCommunityInstances()` was referenced but nonexistent, so I restored it in a subsequent patch.

PR #36, opened 2023-07-03
Fix #26: Show update button on first open of modal

### LINKED ISSUE #26: Version check updates only on second spawn of modal
@artillect 
- Load the page
-  New version is caught but link does not get updated when opening the modal
-  Close the modal and reopen
-  Link styling updates correctly

artillect:
I fixed the issue in #26, and made some visual changes to the button so it looks cleaner and respects the user's theme.

image

I swapped the button's orange color for `--kbin-success-color`, which changes according to the theme. There's also `--kbin-danger-color` (red) as an option. `--kbin-outline` is a similar orange to the one from before, but it doesn't change from theme to theme.


aclist:
That's good, I figured there was something more generic but haven't delved into the themes and was hoping you knew more about kbin's fallback vars. 

Any reason for this jQuery alongside JS?


```
       $('.megamod-version a').text('Install Update: ' + newVersion)
        $('.megamod-version a').attr('href', updateURL);
        $('.megamod-version a').addClass('new');

        versionElement.innerText = 'Install update: ' + newVersion;
        versionElement.setAttribute('href', updateURL);
        versionElement.className = 'new'; 
```

aclist:
I noticed that links look identical to normal text on your theme; is that desirable?

artillect:
> Any reason for this jQuery alongside JS?

Before I moved the version check out of `showSettingsModal()` I had that there to make sure that the change to the button was persistent each time the modal was opened but I think the jquery is unnecessary now

artillect:
> I noticed that links look identical to normal text on your theme; is that desirable?

Yeah, I've noticed that too, it's not a huge issue but it does hide the author link a bit. I also noticed that the modal is completely unreadable on the light theme, with white text on a white background. Probably a good idea to change the text color to whatever color the body/sidebar are, I'll sort that out in another pull request that cleans up the UI a bit and fixes some other issues

aclist:
> Before I moved the version check out of showSettingsModal() I had that there to make sure that the change to the button was persistent each time the modal was opened but I think the jquery is unnecessary now

Alright, why don't you drop that so that we aren't calling the same code twice?

> Yeah, I've noticed that too, it's not a huge issue but it does hide the author link a bit. I also noticed that the modal is completely unreadable on the light theme, with white text on a white background. Probably a good idea to change the text color to whatever color the body/sidebar are, I'll sort that out in another pull request that cleans up the UI a bit and fixes some other issues

May want to look at the footer, too, as the BG is hardcoded. Preferably use an alternating color or just the body color with different opacity.

aclist:
I've set up a `testing` branch which should make it easier to test against remote resources without changing the main branch. Future PRs should be made against testing, which acts as a staging environment prior to merging to main. All of the remote resources are the same, but the URL schema is different, so we can more liberally update the remote resources and test changes.

artillect:
Awesome, that should make things a whole lot easier to keep track of

PR #44, opened 2023-07-05
Remove the language filter since it is built-in to kbin now

artillect:
The most recent kbin update added language filtering functionality, so this mod is no longer needed. Might be a good idea to leave it in until the other instances update as well though


aclist:
Alright, I didn't see too many other topside changes aside from the addition of a thumbnail container for empty thumbnails

Issue #32, opened 2023-07-02
Align fa-bug icon with adjacent link


aclist:
@artillect

aclist:
Resolved by 1873d1890de12696acd355aa4b3054f9f4782409

Issue #23, opened 2023-07-01
Fixed height and y-overflow


aclist:
Need to determine whether:

- Modal should be a fixed height with overflow content therein flowed through a scrollbar
- Modal should just cover the entire screen
  -  Affects dock button, which should be dropped in that case
- To open the entire menu in a new tab

Consider the possibility/likelihood of more verbose information/configs being added in the future and the need to make things feel uncramped.

artillect:
Should be pretty easy to do the first two with a little bit of CSS, I'll get on it. I'm not sure if we'll need to open it in a new tab as long as we make the UI responsive

aclist:
OK, I've added a sticky footer that contains a magazine link and bug report link and should be more conducive to a full-bleed modal.

aclist:
Three workarounds for wanting to preview the results without the modal occluding the page, in ascending order of desirability. 

First, let's assume that we are covering the page and adding a scrollbar.

1. Image tooltips. Hover an icon and you'll see a sample image or gif of what the feature does. Counterpoint: I really don't like this in principle, too fancy and overengineered, need to load external images or base64 encode them, adds overhead with processing the images, and need to load some fancy tooltip popup logic.
2. Lightbulb icon at the top to "dim" the modal contents so you can see through to the parent window. Counterpoint: seems a bit gimmicky, and dimming may cause things to look not exactly as they actually would.
3. Icon at the top to "roll up" the modal into a single navbar (that could still be docked up or down). Counterpoint: not much that I could think of, aside from the need to unfurl it again when done previewing.

So if you want to see the results, the modal gets out of the way and you can see the page clearly, then unroll it again. Thoughts? This is potentially better than a new tab altogether, since at that point we aren't operating on the same window context and it becomes necessarily more complex.

Alternatively, people could just close the modal entirely.

aclist:
Also, (in case it wasn't obvious) the main reason I suggest full screen, fixed height is because it's disorienting to have the modalContent element jump around in size when the content is a different length.

aclist:
I tried setting `.megamod-settings-modal-content` to `height: 100%` and `grid-template-rows: 0.2fr` and it seems to look pretty good in fullscreen and gives much more room to work with! I don't think this is responsive, though, so the main grid items might occlude the header depending on the resolution. Might be good to put the sidebar, body, and helpbox into their own div that stays below the header line. For that matter, I don't think any of the design is very responsive for mobile right now.

artillect:
[...]

PR #45, opened 2023-07-06
Responsive modal, and other UI fixes

artillect:
Here's a list of the big changes so far:

- Modal is now full screen on laptops and smaller devices
- Dock icon hides when full screen
- Modal is centered on large devices to make it more comfortable to read
- Fixed issues with unreadable text in light mode
- Adjusted colors to use theme variables

**On a large device:**
screenshot of the kbin Enhancement Suite settings modal as a popup

**On a medium-sized device:**
screenshot of the kbin Enhancement Suite settings modal as a full screen menu on medium devices

**On a phone:**
screenshot of the kbin Enhancement Suite settings modal on a phone

I've got the fixed height part of #23 done, now I just need to figure out how to get the scrolling to work with all of the nesting (CSS is hard 😤). I also restored the color change when you hover over one of the links in the sidebar or the mods list. I'll tackle #32 and do some more cleanup tomorrow, but I think this is pretty close to being done.


aclist:
Looks great. I'll check the scrolling later and see if it's a simple fix. I assumed you would just scroll the entire body of the modal.

As for the floating modal, if going that route, you'd probably (?) want to restore the lightbox/opacity effect so that it's easier to tell apart relative to the stuff occluded behind it on the page. I originally went with a full bleed "modal" because I don't love the fanciness of floating modals and how they tend to have this goofy effect, so just went with a kind of full width "dock" look. I can see why a modal that is not the width of the page might need an extra readability boost, though.

I'm not really partial to one way or another. I haven't checked out the code yet, so can't say what my reaction would be other than seeing the picture. I'm not the best judge of this because my kbin is so modded at this point that it looks nothing like the original layout.

It looks like we probably need some extra padding between the helpbox toggle and the HR separator. Put it in alignment with the top edge of the other grid items?

My other question is, what does the dock button do now that the modal is centered on the screen?

aclist:
By the way, to reduce human error, the version does not need to be manually bumped anymore. I created an automatic precommit hook to be used by the maintainer. Subatomic commits within a patch shouldn't need to bump the version; it will be bumped once when the patch (regardless of size) is integrated. For that matter, we could potentially drop the VERSION file altogether, since you can use `GM_getResourceText` to read the remote masthead of the `megamod.user.js`and parse it as a text string rather than as JS. I'm not against keeping the VERSION file if it's simpler, although we're just bumping the version in two files when using the git hook.

artillect:
> Looks great. I'll check the scrolling later and see if it's a simple fix. I assumed you would just scroll the entire body of the modal.

Thanks! It might be easier to do than I think, but making auto-sized divs scrollable gets pretty funky sometimes.

> As for the floating modal, if going that route, you'd probably (?) want to restore the lightbox/opacity effect so that it's easier to tell apart relative to the stuff occluded behind it on the page.

Yeah, I've been thinking about that. I think a drop shadow might work well for making it more distinguishable from the contents of the page.

> It looks like we probably need some extra padding between the helpbox toggle and the HR separator. Put it in alignment with the top edge of the other grid items?

Ah, good catch. I'll get them all lined up with each other.

> My other question is, what does the dock button do now that the modal is centered on the screen?

It switches between docking it to the bottom of the screen and the floating modal. I feel like it works pretty well as a replacement

> By the way, to reduce human error, the version does not need to be manually bumped anymore.

Sweet, I was wondering how that worked.

> I'm also going to update the maintainer's precommit hook so that mod authors don't have to update megamod.user.js at all with @requires or touch the funcObj. 

Amazing, that'll really simplify things when testing stuff out and implementing it. I think I can work with those limitations too

I'm gonna try to finish up this PR tonight and give it a polish, hopefully I can sort out all the issues

aclist:
> Yeah, I've been thinking about that. I think a drop shadow might work well for making it more distinguishable from the contents of the page.

Simply try this, should do the trick:

```
.megamod-settings-modal{
     background:rgba(0, 0, 0, 0.5)
}
```

artillect:
Alright, I think this is good to go, not much has changed visually, but I've fixed everything mentioned above, and added some improvements.

- Aligned bug icon with link, and inserted it into link so the whole thing is clickable
- Changed the modal so it doesn't overflow the screen, and made its children scroll instead
- Scroll entire modal on mobile instead of the individual divs on their own
- Adjust layout so everything lines up nicely
- Dark background when undocked, drop shadow when docked
- Decreased height of modal when docked

I might have missed some smaller changes, but that's the gist of it

aclist:
Looks pretty dang good. Take a look at what happens when you click, e.g., the Easy Emoticons entry, though. The longer description causes the width to expand instead of wrapping.

aclist:
Also noticed that if you are in docked mode and toggle between General/Threads, the height adjusts (not fixed).

aclist:
I'd be OK with dropping the docking feature if it seems too fancy/pointless, btw

artillect:
Hmm, I can't get the width to change on my end when I open the Easy Emoticons entry, but I'll see what I can do about it anyways

PR #48, opened 2023-07-07
Prevent modal from changing size

artillect:
Here's a fix for the issues with the modal's height changing when switching pages while docked, and for the width changing when opening a mod with a long description (I think, I couldn't reproduce the problem, but I believe I've prevented it from happening in the future.)

### DIFF
--- a/megamod.css
+++ b/megamod.css
@@ -36,6 +36,7 @@
     border: var(--kbin-options-border);
     position: relative;
     min-width: 1360px;
+    max-width: 1360px;
     min-height: 80vh;
     max-height: 100vh;
     display: grid;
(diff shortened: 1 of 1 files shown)


aclist:
Alright, thank you, will test this in 12 hours. I can take a video if necessary. What browser are you using for testing? I am using Firefox 113.0.2.

artillect:
I'm using Edge 114.0.1823.67, I don't think it's a compatibility issue since I'm not using anything too new, but it might be.

aclist:
Works great now. I'm not completely sold on the docking thing, but otherwise it's perfect. Not sure if we need more padding between scrollbars and each grid section or if that is just my setup that makes them look really tight.

PR #49, opened 2023-07-08
Fix collapsible comments causing infinite loop

artillect:
I believe I've sorted out the issue with the collapsible comments causing an infinite loop with the mutation observer, give this patch a try and let me know if it works


artillect:
Upon further inspection the mutation observer loop doesn't seem to have completely stopped, but it's no longer going crazy whenever you collapse a comment. Maybe something else is causing it to loop too?

aclist:
Hmm, I don't see any recursive issues with it at the moment, seems responsive over here, but haven't looked at the debug output too closely. When are you seeing a loop?

The mutation observer itself doesn't disconnect until page reload; it's technically "live" for the duration we are on the page, since we need to catch ad-hoc events, but the buffer (should be) cleared after each invocation so that we aren't spamming old events from the stack.

artillect:
I couldn't get it to loop, but every once in a while it seems like the mutation observer gets triggered for no good reason. I think it's happening every 30 seconds or so, but it's pretty inconsistent

aclist:
I noticed this as well, but with or without any KES logic. I think kBin itself tries to periodically update its resources (some keepalive type action?), but I may be wrong. This could easily be investigated by running an observer without KES and seeing what the mutations contain.

PR #50, opened 2023-07-08
Add config option to collapsible comments, and prevent it from loading when there are no comments

artillect:
Just a few small tweaks to the collapsible comments mod

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -430,10 +430,10 @@
       {
         "type": "radio",
         "initial": "Headers and bars only",
-        "label": "Collapse by clicking on:",
+        "label": "Collapse by clicking:",
         "key": "click",
         "values" : [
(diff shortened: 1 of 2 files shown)


aclist:
In theory I think it's bad practice to require reload for the option to update. I've been trying to go back and refactor my scripts so that if the styling is already in place but an alternate radio option is selected, it goes back and rewrites them with the desired appearance. Does that seem doable here?

artillect:
It's probably possible, but it touches a lot of things in a lot of places so it might be difficult to clean things up all the way. I'll see what I can do about that

aclist:
Right, didn't notice this was the complicated comments mod. Looks like you have it auto reloading when toggled. If it's too much of a nuisance to change around, I guess that is not too bad.

PR #51, opened 2023-07-10
Easy Emoticon Feature: Added command list modal functionality in version 0.7.8

minnieoo:
This PR adds the functionality to Easy Emoticon of a command panel that can be summoned with /help or /commands, which conveniently allows you to scroll through the list of commands on Kbin through a modal, rather than rely on memory or an external document like before.


aclist:
Would you mind making this patch against the `testing` branch (`git checkout testing`) and resubmitting? Thanks

PR #55, opened 2023-07-10
feat: added kibby svg to /images

minnieoo:
added kibby svg to /images. replace folder logo with kibby, everything else is the same


aclist:
Working great! You can test it in the latest version using the 'Toggle logo' feature. I think it would be good to add a white and black stroke to the outer edge, though. This way if it's on a dark background, details like the beak won't get lost, and likewise if it's on a light background, the black stroke should help outline it. This is what is usually done with subtitles to ensure they are readable on any BG color. Thoughts? Here's what it currently looks like on mine:

a

minnieoo:
I had the exact same thought lol. i thought perhaps I should at least have a white stroke on the beak at least for readability, but i didnt end up going for it since it was a bit better when i turned my userstyle off. but yea, it would be better to had it more readable for all color bg's. 

as for the double stroke, do you mean a black stroke outside of the gradient stroke for the body?

aclist:
Yeah, I didn't explain it correctly, but subtitles are usually white text with a black stroke. This way if the BG is black, it will still show, and if the BG is white, it'll be outlined. That's the basic idea.

I don't know if you necessarily need a stroke all around everything, but since the body is white and the beak is black, it's always going to be one or the other that is not perfectly visible if the page BG is either light/dark.

I think it's a good idea, since someone's theme could be any color. But I'll leave the implementation up to you

aclist:
White is a bit better, although the kbin text disappears (that's a design flaw in the original kbin logo)

aclist:
Middle one looks pretty great, or am I crazy? It's subtle but probably just enough

minnieoo:
ah okay, whatever works. ill add that one in

aclist:
I think I see what you mean after testing it live...this is tricky. It looks decent at higher scaling levels, but looks kind of crunchy at <120% zoom level. Not sure what is a good approach here other than putting the whole icon onto a background?

minnieoo:
yea, it is tricky. i can try the idea's of playing it on perhaps a gradient bg and seeing what it looks like, ill update you soon enough

minnieoo:
@aclist i tried a circular gradient and it worked out better visually than i expected. they might be a bit too close together, the text and the icon, but i think it looks solid even from far away, if needed i can push them apart. also it might be a bit too big compared to the text? the original folder icon touched the bottom and top of the canvas so i just followed that, but i can scale it if need be as well.

im gonna attach a screenshot and also the svg so you can quickly test it yourself in the dom like i did. lmk what you think of this direction. the svg for some reason appears black when u click it, but its not i swear lol, the link is good to use.
image
image

kbin_logo_kibby11

aclist:
Looks good! Let's go with that. I agree they look a bit close, don't think there would be any problems if you space them as long as the total canvas dimensions of the image are the same

aclist:
I added the image in over here, so you don't need to make an additional PR. It is looking swanky

minnieoo:
oh good, glad you like it! sorry for my late response! i was sleeping lol. i dont know if you still wanted me to space them apart but here it is:
kbin_logo_kibby_2

minnieoo:
btw @aclist , i realized my script says login required when it doesn't require login, i must have missed that setting and didn't correctly label it. would i submit a pr for that small correction?
image

aclist:
I actually labeled it this way myself, because I couldn't find any text fields on the site that you can use your mod on except replies/messages. It seems it only targets reply fields, correct? I tried using it in search boxes but it did nothing. Is there any way to insert emoticons without logging in? You can't reply to comments if you don't log in. Am I wrong?

I agree, `textarea` is kind of the technical term. Will fix both of those for you

minnieoo:
omg i wasnt thinking straight, youre not wrong lol. idk why i didnt even consider that 🤦🏽‍♀️🤦🏽‍♀️ yea it only works on nearly every text input field, whether it be comment boxes, profile bios, dms, etc. ignore me lmfao

PR #57, opened 2023-07-10
WIP: Helping out with the docs

artillect:
Here's what I've got so far, I mostly just filled out the "usage" section, but I made a few changes in earlier sections too.

Is it possible to have one file for both the light and dark mode? I've never used asciidoc before so I have no idea, but it seems like the sort of thing it's built to do based on what little I've read about it

### DIFF
--- a/docs/kes.adoc
+++ b/docs/kes.adoc
@@ -20,9 +20,9 @@ It also provides a simple, human-readable format for script authors to integrate
 KES is itself a userscript that bundles other scripts. As such, it requires any of the below 
 browser extensions in the *monkey family:
 
-- TamperMonkey
-- GreaseMonkey
-- ViolentMonkey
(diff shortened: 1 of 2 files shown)


aclist:
Thanks for the updates.

> Is it possible to have one file for both the light and dark mode? I've never used asciidoc before so I have no idea, but it seems like the sort of thing it's built to do based on what little I've read about it

ADOC files are flat HTML files. Without a JS toggle, it's not possible. There is a plugins API, never looked at it. I believe the strength lies in the portability of the file and that the template will produce expected results (like a (La)TeX  file). You can add a script after the HTML is generated, but it becomes another file to keep track of.

It does support variable expansion, which is what I use when updating the `Last updated {d}` string, so you could script this to generate two versions, too. I usually just copy over the diffs.

PR #64, opened 2023-07-12
Fix compatibility with Violentmonkey

### LINKED ISSUE #63: [BUG] KES doesn't load properly in Violentmonkey
**Describe the bug**
After installing KES with Violentmonkey, it throws a syntax error in the console, and it doesn't load at all. Removing the offending `await` on line 133 allows the script to load, but none of the options appear in the settings

**To Reproduce**
1. Install KES with Violentmonkey
2. Open any kbin instance
3. Check the console for errors

**Expected behavior**
It should load the settings menu properly and show all of the options

**Screenshots**
image

**KES version:**
- 1.1.1

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Firefox
 - Browser version: 114.0.2

artillect:
I removed the (apparently nonfunctional) await on line 133, and replaced all of the `GM.` functions with `GM_` functions, as violentmonkey doesn't have those built in

Fixes #63

### DIFF
--- a/kes.user.js
+++ b/kes.user.js
@@ -14,8 +14,8 @@
 // @grant         GM_getResourceText
 // @grant         GM_xmlhttpRequest
 // @grant         GM_info
-// @grant         GM.getValue
-// @grant         GM.setValue
+// @grant         GM_getValue
+// @grant         GM_setValue
 // @grant         GM_getResourceText
 // @grant         GM_setClipboard
 // @connect       raw.githubusercontent.com
@@ -125,12 +125,12 @@ function makeArr(response) {
     var content = response.responseText
     const jarr = JSON.parse(content)
     //TODO: wait on promise and set warning string if unreachable
-    GM.setValue("json", jarr);
+    GM_setValue("json", jarr);
 };
 
 fetchManifest();
 checkVersion();
-var json = await GM.getValue("json");
+var json = GM_getValue("json");
 var css = GM_getResourceText("kes_css");
 GM_addStyle(css);
 


aclist:
Violentmonkey seems to be strict about that; I was picking throuhg it and able to resolve that part by wrapping it in an async function:

```
async function getJson(){
   let json = await GM_getValue("json");
  return json
}
const j = getJson();
j.then((result) => {
  //process result here
})
```

But applying your fix above seems to clear the first issue, but I'm getting issues with `getModSettings() is not defined`. Are you seeing that?

aclist:
To clarify my comment, this allows the JSON to be processed, but I don't think any of the options function because `getModSettings()` is failing

aclist:
Seems to occur when any of the mods calls `getModSettings()` from its scope, as if they aren't operating in the same scope as the main script. Could be a violent monkey setting on my side?

artillect:
Hmm, I'm not seeing any settings that'd affect that unless I'm missing something

aclist:
Have a look at this:

https://github.com/violentmonkey/violentmonkey/issues/173
https://github.com/violentmonkey/violentmonkey/issues/408
https://violentmonkey.github.io/posts/inject-into-context/

I tried violentmonkey's various injection modes, but none of them worked unless explicitly doing this:

OLD:
```
function getModSettings()
//...
```

NEW: 
```
unsafeWindow.getModSettings = function(namespace){
//...
```

Second method works, all settings load. However, I don't remember all of the implications of this, so don't think it's a good idea to include this in the normal script.

To better identify the problem surface: does this only affect VM + Firefox? Does the same script in VM + e.g. Chrome work when using VM's "auto" page injection context? (Can you test that?)

It seems related to Firefox's stricter CSP policy.

artillect:
I get the same issue in Chrome with VM.

Is there a way to pass the function to the mods so they have access to it in their namespace? This is a bit beyond my JS knowledge

aclist:
I think this should fix it and is a fairly vanilla approach:

```
window.getModSettings = function(namespace){
  let settings = localStorage.getItem(namespace);
    if (!settings) {
        settings = {};
    } else {
        settings = JSON.parse(settings);
    }
    return settings;
}
```

Should bind it to the global window context and allow this one function to be seen by the scope of the others.

I think the "unsafewindow" is a bit of a misnomer in that it's widely supported and modern browser should already be isolating injected JS from self JS. I think it used to be the case that pages could "see" a third party script this way, but that should be ancient history by now sandboxing is already taking place.

Some additional discussion here: https://stackoverflow.com/questions/65552070/why-do-userscript-managers-still-suport-the-use-of-unsafewindow

aclist:
I have pushed the fix, please have a look.

Also, looks like your collapsed comments mod is calling `GM.addStyle` instead of `GM_addStyle`, and this is causing VM to exhibit problems. Would you mind taking a crack at that and pushing a new PR?

artillect:
I'll fix that real quick, we should add a note in the documentation to only use `GM_` functions for better compatibility across the `*monkey` family

artillect:
Looks like your fix is working

aclist:
> I'll fix that real quick, we should add a note in the documentation to only use GM_ functions for better compatibility across the *monkey family

That makes sense, probably suggest TM over VM and GM just because it's the most feature complete and mature at this point, but that's a great idea, will look into adding it to a testing suite that will prevent these kinds of things from inadvertently making their way in.

PR #68, opened 2023-07-12
Fix collapsible comments compatibility with violentmonkey

artillect:
This should sort out the issue with collapsible comments not working with VM


aclist:
https://i.imgur.com/X2yAHnH.jpg

PR #81, opened 2023-07-13
Adds safeGM function for generic support of *monkey variants (#74)

aclist:
See commit notes for details.

Needs testing on combinations of:

- Firefox
- Chromium
- Other

- GreaseMonkey
- TamperMonkey
- ViolentMonkey
- Userscripts

Main affected problem surface based on reports is:
- Firefox + GreaseMonkey
- Safari + Userscripts

PR #82, opened 2023-07-13
Safegm breakout

Oricul:
This breakout fix will allow mods to access safeGM.


Oricul:
This should be considered before merging https://github.com/aclist/kbin-kes/pull/81 into testing.

Issue #47, opened 2023-07-06
Chore: update precommit hook to insert URLs and mod names


aclist:
Update the maintainer's precommit hook so that mod authors don't have to update megamod.user.js at all with `@requires `or touch the funcObj. The author can just:

- Test the mod locally using whatever URLs they want for the source mod

When making the PR:
- Update the manifest.json 
- Add the mod to the mods directory
- That's it.

The hook will be responsible for updating megamod.user.js with the correct path to the resource and updating the funcObj. This should reduce the problem of having the wrong URLs or names making their way into the PR and ensure that after the mod is tested and approved, the resources are updated programmatically, along with the version bump, for one unified change.

- Add this to documentation
- Optional: for maintainability, consider parameterizing and regenerating the entire header based on the manifest of mods to flexibly support URL/branch changes/supported GM API calls

aclist:
Resolved by 85060d43fdc7f12a02d6e9a04c1248dedfbff9ee

PR #84, opened 2023-07-13
Add notifications panel userscript

honkaiimpact3rd:
(＾＾)ｂ


Oricul:
Is it worth adding to the description regarding not supporting Firefox (and flavors) out of the box?

honkaiimpact3rd:
No need to, updated it and it's no longer required. There's still a part in css that uses `:has()` but it's a small cosmetic change that isn't important.

aclist:
Glad you got it sorted, users will be happy to have this one included. Will review this ASAP tomorrow

Oricul:
I've found some issues with this implementation, but it's not far off. 

1. The primary issue https://github.com/aclist/kbin-kes/commit/3d1d936ed9906db6d545ef54e3d2be7de601bb2f - The commit makes it look uglier than it is. My findings: With the current implementation it's wrapped in a function which makes everything uncallable. There was also no way to toggle off and was always running, even when toggled off in settings. Needed to add the toggle handler and subsequent shutdown function. Was also an issue where it was being triggered over and over - I believe this to be potentially a KES issue when recurs is false, but for this instance, I handled with a simple check.
2. These two are lesser issues which can be addressed by @aclist - https://github.com/aclist/kbin-kes/commit/b411b44dde2b0afac0e5ec5cfb9edc5e3cfa94d2 and https://github.com/aclist/kbin-kes/commit/3459f207781c78b1e60b3f8f5496c1abc3fd8a5e

Hope this is useful.

aclist:
> These two are lesser issues which can be addressed by @aclist - https://github.com/aclist/kbin-kes/commit/b411b44dde2b0afac0e5ec5cfb9edc5e3cfa94d2 and https://github.com/aclist/kbin-kes/commit/3459f207781c78b1e60b3f8f5496c1abc3fd8a5e

This is not an issue per se, once the tooling described in https://github.com/aclist/kbin-kes/issues/47  is complete, it stops being the repsonsibility of mod authors to include those files, so @honkaiimpact3rd  did it correctly based on the spec in the documentation. Oftentimes we forget to include or add a typo in the mod name declaration, so would prefer to do it programmatically rather than manual insertion.

Oricul:
> > I believe this to be potentially a KES issue when recurs is false, but for this instance, I handled with a simple check.
> Can you elaborate on this?
Recursion is turned off for this mod in the manifest, but it's entrypoint was being triggered every 10-20 seconds it seemed. There's also a few events in this mod that I didn't dig into yet, but I didn't think this mod would be triggering it's own entrypoint - so I just added a check to see if it's already present. 

Unfortunately, I didn't dig more into it yet.

aclist:
Presumably it's related to the load event listener, which you removed. I was just wondering if there was something with KES itself that is erroneously triggering, but I don't think that should be the case.

aclist:
Everything checks out and it's up to date with the current code. Is it customary for the notifications panel to take a sec to load in the content? Please take a look on your end after updating your copy of the testing branch.

Issue #72, opened 2023-07-12
[MOD] Change comment/thread font size


artillect:
**Is this request to port an existing userscript or mod, or create a new one?**
Link to the source script if applicable.

**Describe the solution you'd like**
Add the ability to adjust font sizes for threads and comments independent of the page's font size.

**Additional context**
For some users, the font size for comments/threads is too small, and increasing the font size impacts how the page is laid out on mobile. This comment is where this idea is from: https://kbin.social/m/kbinMeta/t/183928/-/comment/741140

aclist:
One thing we should start looking to do is consolidating features so that we don't have 1,000s of mods that, e.g., hide one element. It seems like this would fit well into a mod that controls a variety of different onscreen elements, and you can granularly adjust the ones you want. Or turn it off altogether.

```
Author: you
Dynamically changes the font size of elements

Headers [dropdown]
Comments [dropdown]
Post description [dropdown]
Usernames [dropdown]
```

etc

artillect:
I was just thinking about that, I think it's a good idea. Speaking of organizing things, it might be nice to sort some of the stuff from the general page into more specific pages so the options are a little more spread out. I think adding a few extra pages could be good once we add some more mods, like "Navigation" for example

aclist:
Yes, although we can't do that too much after a certain point, as it is liable to be confusing going forward, once people get accustomed to certain mods being in a certain place

aclist:
Closing as duplicate of https://github.com/aclist/kbin-kes/pull/171

PR #85, opened 2023-07-13
Change Font Size Mod

minnieoo:
This is a customizable mod that allows the user to change the font size of posts, comments, or both simultaneously. 

https://github.com/aclist/kbin-kes/assets/137783738/a436c923-e398-4a6f-9bde-f1bca4b09c7e


aclist:
Take a look at this: https://github.com/aclist/kbin-kes/issues/72

I think it makes sense to incorporate as many elements as possible here, such as headers, posts, names, comments, description.

I would also take a look at the number input type: https://developer.mozilla.org/en/docs/Web/HTML/Element/input/number and map it to a point size increment.

This gives the user very granular control over their preferred point size and you don't have to set predefined sizes ahead of time, plus you can arrow up/down and increment/decrement on the fly. Maybe start it at a default size between 12pt or whatever is the equivalent to the default size in rem

minnieoo:
o i missed that discussion, thanks for letting me know. ill try to apply this 👍🏽

minnieoo:
@aclist maybe im wrong but is there no way to add steps for the number input within the manifest? for now i was just gonna add them using the dom but i thought maybe this would be good to have available. i didnt see anything for it in kes.user.js

aclist:
We didn't implement steps, now that you mention it :crying_cat_face: 

Nor min/max... Number input seems like one of the more common ones to be used, so we should definitely extend this

aclist:
I am adding this now, should be quick to do.

aclist:
Added in 2.0.0-rc.25, just tested the helpbox and increments appearing in localstorage, didn't test it end to end, but should be good to go

Issue #66, opened 2023-07-12
[FEAT] Multiple Author Support


Oricul:
**Is your feature request related to a problem? Please describe.**
Some scripts may have multiple authors, or be abandoned and picked up by someone else.

**Describe the solution you'd like**
Support an author array in manifest.

**Describe alternatives you've considered**
I've thought about the description field, but it would make things too bulky for users I think.

**Additional context**
![Example](https://files.catbox.moe/qfn7vd.png)

aclist:
Sounds great and easy to implement, agree that we shouldn't be stuffing the description field too much

PR #86, opened 2023-07-13
Feature: Multiple author support

Oricul:
While implementing multi-author support (in an array), I changed some nearby items. I broke HTML strings into their JS counterparts to be more readable. I also made author links link to the current instance instead of kbin.social always.

### DIFF
--- a/kes.user.js
+++ b/kes.user.js
@@ -247,18 +247,58 @@ function showSettingsModal() {
     }
     sidebar.appendChild(sidebarUl);
 
+    function kesFormatAuthorUrl(author) {
+        if (author.includes('@')) {
+            return 'https://' + window.location.hostname + '/u/' + author;
+        } else {
+            return 'https://' + window.location.hostname + '/u/@' + author + '@kbin.social';
+        }
+    }
+
     function openHelpBox(it) {
         const settings = getSettings();
-        settings.lastPage = it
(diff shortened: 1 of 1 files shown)


Oricul:
This PR is to implement https://github.com/aclist/kbin-kes/issues/66 .

aclist:
Very nice, the logic to add elements was unspeakably convoluted and getting unreadable. I keep thinking there is more room to abstract here somewhere by creating smaller functions that construct the elements in a less imperative manner, but it doesn't seem important for now because the UI is mostly fixed now. It would make future changes easier, though

aclist:
Would you mind resubmitting this against `testing` now that the safeGM patch is merged in? I thought I resolved the merge conflict, but ended up killing some chunk of code in the process, so had to revert it. Sorry.

Oricul:
> Would you mind resubmitting this against `testing` now that the safeGM patch is merged in? I thought I resolved the merge conflict, but ended up killing some chunk of code in the process, so had to revert it. Sorry.

Unfortunately I can't.
0 changed files

aclist:
> Can you revert the change in this PR?

Unfortunately I cannot, the content is already stale. If you have a local version of the multi author code that differs from the current state of testing, create a new branch and PR?

PR #92, opened 2023-07-14
Patch: better async handling when remapping GM API calls

aclist:
Working in TM, GM, VM. Notes:

- `event.stopPropagation()`  does not work on GM. Not sure what purpose it was serving on line 511, though.
- When selecting pages, the CSS rules responsible for keeping them at fixed height seems to not be applying (might be FF only)
- Position of wrench icon does not seem correct

Hypothesis: CSS is getting partially truncated on insertion?


aclist:
Think I see the issue, accidentally clobbered the `@resource`s in the header even though they are still being used by the `GM_` variant

PR #97, opened 2023-07-14
Multi author support fix

Oricul:
Untested, but it should be the same as before.


aclist:
Thanks, seems to have no problems over here.

Issue #90, opened 2023-07-14
[BUG] First load: open settings = undefined json


Oricul:
**Describe the bug**
KES settings will appear (mostly) empty, and there are errors in console regarding undefined json.

**To Reproduce**
1. Remove KES' "Externals" and "Storage".
2. Visit a supported kbin instance.
3. Click KES' wrench icon.
4. Settings will be (mostly) empty.
5. Open console and you will see messages about undefined json.
6. Refresh the page (or navigate to a thread, etc.)
7. Open KES settings, and content will be present.

**Expected behavior**
Opening KES settings, even on first load, should load settings properly.

**Screenshots**
![Example](https://files.catbox.moe/vks0g5.png)

**KES version:**
- Tested in 2.0.3, but I would almost guarantee it's present prior to this too.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: LibreWolf (Firefox)
 - Browser version: 115.0.1-1
 - UserScript Manager: TamperMonkey

aclist:
This is probably going to be fixed with the new patch, since the original safeGM implementation just naively remapped API calls to different names without actually handling async

Oricul:
Good to hear. Let me know when testing branch has this fix and I'll confirm. It currently does not.

aclist:
I'm just sitting down to untangle whatever it was I did yesterday, will update after the tea kicks in.

aclist:
The patch is here: https://github.com/aclist/kbin-kes/pull/92

Unfortunately, some problems with TM, but it works on the other two. You'll also need to host the new safeGM somewhere and change the URL, since it's not currently merged in.

Oricul:
It's merged on my fork/branch. I'll take a look to see what I can find.

Oricul:
@artillect - What are you seeing as not working in TM?

aclist:
Think you meant to reply to me? It was not loading in TM on the initial commit, but it was because I dropped the external resources from the header. I submitted a new change and it's working on TM as well, but something is awry with the wrench icon.

aclist:
![a](https://github.com/aclist/kbin-kes/assets/92275929/9fadffc9-3aca-4649-9043-64021324ba52)

aclist:
Regarding the initial bug, with this patch, tried deleting the externals from TM and force reloading, and the UI loads.

Oricul:
[...]

PR #106, opened 2023-07-14
fix: update remaining paths for helpers

Oricul:
Fixes https://github.com/aclist/kbin-kes/issues/90#issuecomment-1636403888

Note: My VSC likes to fix formatting. Sorry!


aclist:
> Note: My VSC likes to fix formatting. Sorry!

That's fine, I haven't been running things through a linter much except in between major versions, and even then, I was doing it sporadically. We should probably agree on a standard for indentation at least. I was going with four spaces.

Oricul:
My tab spacing is set to 4 spaces as well.

PR #107, opened 2023-07-14
feature: respect kbin style options

Oricul:
Will now respect kbin style selection (i.e., square or rounded borders). Code refactor and change to support safeGM because why not?


aclist:
Personal trivia: I hate rounded corners :smile_cat:

Issue #87, opened 2023-07-13
[FEAT] Options shown based on other options


Oricul:
**Is your feature request related to a problem? Please describe.**
One thing that **kbin-mod-options** allows for is for some options to be hidden or displayed based on what another option is set to. I'd like to see similar functionality added here. 

**Describe the solution you'd like**
Option fields would need expanded with 2 additional field options, I believe. Perhaps **depends_on** and **depends_value**? This could also be expanded to _only_ show settings if the mod itself is enabled to begin with. The event listener would have to re-render the hBox as well to show updates - shouldn't be too difficult as it's already listening.

**Describe alternatives you've considered**
Current alternative right now is to have settings visible that aren't actually active.

artillect:
Instead of hiding them completely, it could be nice to have them be disabled until the appropriate option is enabled, so they still appear, just greyed out

Oricul:
Why not both with a **disabled_state** that defaults to simply being disabled?

aclist:
For the sake of getting the initial prototype up and running, I thought it would be fast enough to just include the options inline, since they aren't liable to really clash with each other much and it's the user's responsibility.

I did consider the contextual toggle during development, but concluded it would be a waste of time at that point.

I'm in favor of this change, provided we can nail down a simple, reliable spec that doesn't overload the manifest with too many extraneous options or bloat the menu with overengineered fancy effects.

I also agree that hiding settings is a bad pattern (feels like mobile), as what a thing does should for the most part be visible from the start without having to "unlock" things.

Pamasich:
Why was the linked pull request on this closed?

PR #109, opened 2023-07-15
feature: disabled and depends_on settings

Oricul:
For: https://github.com/aclist/kbin-kes/issues/87


kes.user.js
```diff
@@ -427,6 +445,7 @@ function constructMenu(rawJSON, rawLayout, rawCSS) {
                     let label = document.createElement('p');
                     label.className = "kes-field-label";
                     label.innerText = json[it].fields[i].label;
+                    kesCheckSettingsOptionDisabled(label, json[it].fields[i].depends_value, getModSettings(json[it].namespace)[json[it].fields[i].depends_on], getModSettings(json[it].namespace).state);
```
aclist, line 448: This seems unnecessarily verbose and I didn't get why it makes two calls to getModSettings when the contents of the object have already been set on line 439. We already have the entire contents of the settings for that namespace (ns), so no need to call the function again. Before we enter the switch statement, the contents of the settings object are parsed out with those rows of lets to make them available through the entirety of the switch statement. The only thing that changes is the field type, everything else remains the same. (Unless I'm misunderstanding)

aclist:
Hi, would you mind checking your inbox at kbin social? Thank you! :bow:

Issue #142, opened 2023-07-18
[MOD] Collapsible comments: integrate with scripts using endless comment scroll


aclist:
"Like the collapsible comments didn't work with the endless comment scrolling from kbin usability pack and I found the ones from kbin enhancement script just had more usable features that come with it, like collapsing replies and a default collapsed state."

Reposted from kbin:
https://kbin.social/m/kbinMeta/t/183928/kbin-Enhancement-Suite-a-community-curated-script-manager-that-lets-you#entry-comment-914002

- Note: infinite scroll would be good to support OOTB and elaborate more generally so that it can be used as a drop-in utility function in other mods

Issue #130, opened 2023-07-16
[BUG] Improved collapsible comments: after 10th reply, nesting breaks


aclist:
- Create a reply chain 9 items deep, then post a 10th reply and thereafter
- The nesting breaks and restarts underneath the old chain. (Intentional?)

artillect:
I've run into this a few times before, but I'm not sure what causes it though since those comments don't appear to be any different from 9th-level and below. I'll take a closer look and see if I can get to the bottom of it

aclist:
Resolved by https://github.com/aclist/kbin-kes/pull/144

Issue #78, opened 2023-07-13
[BUG] Collapsible comments: last reply in a nested thread is missing expando line


honkaiimpact3rd:
**Describe the bug**
When you add a new reply to something, a thread line isn't made (but it is made in [the previous separate version of the userscript before KES](https://greasyfork.org/scripts/468923-kbin-improved-collapsible-comments)). But it does appear after refreshing the tab. 

**To Reproduce**
1. Go to any thread
2. Reply to an existing comment

**Expected behavior**
A thread line to made

**Screenshots**
New reply doesn't have a thread line
![Screenshot_20230713_132922](https://github.com/aclist/kbin-kes/assets/107812489/c356187d-ac43-4708-8315-9cedc2a191f1)
![Screenshot_20230713_133536](https://github.com/aclist/kbin-kes/assets/107812489/d71852a7-0bfc-4bf0-9c25-c5b83afda6db)

But while using the other, separate version of the userscript, the thread line is here
![Screenshot_20230713_133006](https://github.com/aclist/kbin-kes/assets/107812489/266b5a71-7119-49cd-a448-2ad8202faedc)
![Screenshot_20230713_133734](https://github.com/aclist/kbin-kes/assets/107812489/768ae546-ffa6-4cb3-8ee5-2e2663dbe00e)

**KES version:**
- Version number 1.1.4

**Desktop (please complete the following information):**
 - OS: Fedora
 - Browser: Vivaldi (Chromium)

aclist:
Upon further investigation, the issue is here: 

```
    if (toggle) {
        if (document.querySelector('.entry-comment.nested') || !document.querySelector('.comments')) {
            return;
        }
```

The first branch of this condition is TRUE when submitting a new reply, so it returns from the toggle function without reapplying. But it seems like this logic is necessarily needed to handle collapsing/etc.

aclist:
Resolved by https://github.com/aclist/kbin-kes/pull/144

PR #144, opened 2023-07-18
Fixes and stability improvements for collapsible comments

artillect:
- Show moderation panel under comments with collapsible comments enabled (doesn't work perfectly for top-level comments though, it gets inserted under all of the comment's children)
- Fixes the issue with being unable to collapse new replies you make, this seems to also fix the incompatibility issue with KUP's infinite scrolling comments (I know it uses a separate mutation observer, but I couldn't figure out how to get it to behave nicely with the one built in to KES, it kept infinitely looping like before)
- Fixes the issue with comments not nesting at level 10 and above (regex is hard 🥲 `/comment-level--(\d)/` -> `/comment-level--(\d+)/` was the fix)

This should fix all of these: #142, #130, #78


aclist:
What are the implications of this extra mutation observer here, anything we need to be aware of?

artillect:
> I thought it was something wrong with the arrays

So did I lmao, spent so much time pulling my hair out trying to figure that out when it was such a simple fix

> What are the implications of this extra mutation observer here, anything we need to be aware of?

Honestly, I'm not entirely sure. It doesn't seem to be conflicting with the one in KES. I guess there would be a little bit of a performance impact, I'll take a look at what you wrote for the main observer to see how much I can filter it down.

I'm not sure how to fix the expando line disappearing when you edit a comment. It appears that, when you make an edit, the entire comment div (including the meta, buttons, and children) gets replaced with the edited one. I have no clue how to do it, but it seems like I need to copy the children to a variable when you click the "edit comment" button

aclist:
> I'm not sure how to fix the expando line disappearing when you edit a comment. It appears that, when you make an edit, the entire comment div (including the meta, buttons, and children) gets replaced with the edited one. I have no clue how to do it, but it seems like I need to copy the children to a variable when you click the "edit comment" button and then add them and the buttons back after it gets modified.

The initial report said that it does work in the standalone version, is that the case?

artillect:
I don't think the initial report mentioned editing at all, I'm pretty sure that has always been a problem with my script

aclist:
Good point, I think it only mentions replies. I thought they were one and the same.

aclist:
Oh, so you disabled the recurs flag and you set your own observer, eh? Naughty :D

artillect:
Exactly hahaha, that's what I was using before I disabled it when you added the global observer. I think the issue comes down to how I set it up originally (basically with a separate entry point for my own observer), and I can't figure out how to unravel it.

aclist:
Yeah, I tried to debug it a few times but it just made my head hurt. I don't know what it is, the functions are relatively compact, something to do with the repetitive variable names and the previous child, next child, previous element, next element,  and so on. But I guess that's par for the course when reading someone else's code.

aclist:
So if I understood correctly, the KES root observer triggers when the reply-to field is opened, but after making the post itself, it doesn't subsequently trigger ICC, because we aren't checking the subtree, so ICC sets its own observer? 

This seems reasonable enough for now, not sure what other implications there might be. I think we should treat it as tech-debt and endeavor to revamp this in the future so that mods by necessity don't use their own observers unless explicitly necessary, as it might introduce unwanted recursions down the line.

aclist:
I got it! Finally. I wanted to keep chasing this down because I figured it would improve the core KES functionality and I figured it if affected your mod, it could affect others. Here's what I think we do:

Pass through mutation event targets into the applied mod so that the mod can "see" the mutation and filter it accordingly! The ICC mod was failing because on subsequent mutations, it was returning too early because it "already found" nested comments, but wasn't able to inspect the latest comment and validate it as un-nested, so would quit before reaching `applyNewPosts()`. So your solution was to restore the observer so that after this failed mutation check, the script would spawn its own observer to catch these events.

New logic: mods don't need to set up their own observers because KES passes the entire mutation target into the mod as the second argument to the init function IF the mod is set up as `recurs == true`.

So now you get the exact mutation right at the start of the script and you can demonstrably test whether it matches the type needed for some logic to apply:

```
function initCollapsibleComments (toggle, mutation) {
    console.log("MUTATION IN ICC IS:", mutation)
    const mutClass = mutation.target.className
    if (toggle) {
        if (mutClass === ".entry-comment.nested") {
            return;
        }
       applyToNewPosts()
     }
/// etc...
```
        
Sorry to belabor this point, but I think it will make KES itself more robust and flexible. I'll make the changes upstream later and we should take another look at this PR to refactor it slightly, but I think it's gonna work well.

aclist:
Alright, take a look at this fix. This passes mutation events into the mod and filters them accordingly. Handles:

- Nesting comments on page reload
- Nesting nth comment when replying
- Toggling nesting on/off in KES without having to reload the page

The only thing I didn't check was infinite scrolling, since I'm not sure where to enable that on kbin.

Should be seamless now, no extra observers necessary.

aclist:
Just tested and confirmed with infinite scrolling, it works as well. The only thing I was unclear on was what the below refers to: 

> Show moderation panel under comments with collapsible comments enabled

What is the moderation panel?

PR #148, opened 2023-07-18
Prevent OP label from showing on children of OP when collapsible comments is enabled

artillect:
I made the CSS selector a bit stricter to prevent it from applying the tag to replies to OP's comments

### DIFF
--- a/mods/label.user.js
+++ b/mods/label.user.js
@@ -4,7 +4,7 @@ function labelOp (toggle) {
         let fg = settings["fgcolor"];
         let bg = settings["bgcolor"];
         safeGM('addStyle', `
-            blockquote.author a.user-inline::after {
+            blockquote.author > header > a.user-inline::after {
                 content: 'OP';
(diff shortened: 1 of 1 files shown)


aclist:
Just moved this to be a PR against `testing`, rather than main. 

Basic structure is:

`testing` -> dev sandbox

Maintainer consolidates patches into significant release and then creates `release/<num>` branch. Merge latest stable testing version into here, maintainer pushes as PR, re-review and test the content before merging to `main` (hot branch)

There are a few things in this workflow I want to optimize, but need to get my thoughts down on paper first. Mainly the tag generation post-commit hook is a bit flaky right now, but would like to have standardized tags going forward so that contributors can check out to any version at a point in time without sifting through commit hashes.

artillect:
Ah, whoops, sorry for missing that. I thought I had made it against `testing`, I'll make sure to double check that before submitting in the future.

PR #157, opened 2023-07-19
[Navbar icons as Text] Added font and font weight customization

minnieoo:
I found the original font not very appealing, so this addition enables customization to both the font and the font weight.

Before: 
image
Now:
image
image
image
and many more fonts available

Video:

https://github.com/aclist/kbin-kes/assets/137783738/e189508e-47c5-47cd-9611-4e1e02f373c2

### DIFF
--- a/helpers/manifest.json
+++ b/helpers/manifest.json
@@ -527,6 +527,37 @@
     "namespace": "nav_icons",
     "fields": [
       {
+        "type": "select",
+        "initial": "Segoe UI",
+        "key": "font",
+        "label": "Select a font",
+        "values": [
+          "Segoe UI",
+          "Quicksand",
+          "Calibri",
+          "Arial",
+          "Times New Roman",
+          "Helvetica",
+          "DM Serif Display",
+          "Lato",
+          "Roboto",
+          "Anton",
+          "Poppins",
+          "Bebas Neue",
+          "Oswald"
+        ]
+      },
+      {
+        "type": "range",
+        "initial": 5,
+        "key": "fontWeight",
+        "label": "Select a font weight",
+        "step": 1,
+        "min": 3,
+        "max": 9,
+        "show_value": true
+      },
+      {
         "type": "text",
         "initial": "Search",
         "key": "search",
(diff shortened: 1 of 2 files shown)


aclist:
I'm not sure if all of these fonts are supported OOTB on all systems. Is there a more generic collection we can use?

I didn't even notice the default font was so hideous because I had my browser set to ignore sites' font settings.

aclist:
Tested it and it works on some fonts. Couldn't we do something more generic here like set the font-family property to `sans-serif` so that it respects the user's browser defaults without having the icon font look different from the rest of the page?

What if we drop the font face selection and keep the weight slider?

aclist:
Added in 34c0ac25f9b559d3aa7e2a13f0fb6d7961291c34

minnieoo:
perfect! even better solution :D

minnieoo:
@aclist i noticed an issue. the icons don't seem to be reverting back to the initial font family it was set at (Font Awesome 6 Free) despite the script saying "initial". I got around this by doing  `font-family: "Font Awesome 6 Free"`. not sure why it's not working


https://github.com/aclist/kbin-kes/assets/137783738/beeb0857-470f-43eb-bcb5-4c187bc6a254

aclist:
Good point, let's fix that, and that way we can update the font variable too.

minnieoo:
@aclist random question, what in KES is causing rounded profiles? I am updating my random kibby profile pic for default avatars script and i notice when I have KES on, the pfps are rounded but obviously do not select for the new kibby profile pictures so they remain large and square. also, if i have random little questions like this, where should i ask them? do they warrant an entire discussion topic or issue? not sure where to put these lol

aclist:
I enabled everything to test: looks like it is caused by collapsible comments. Can you make a bug ticket about that and ping artillect?

Doesn't matter where you ask in principle, although there's a label you can use when making a ticket:

a

minnieoo:
I found a much better way to deal with it that is actually the right away, so now its working smoothly even with the rounded pfps and his script selects for it now. though, maybe i should still make the feature req, some people may not prefer rounded pfps, it might be smarter to have it as its own toggle/mod. what do you think?

aclist:
I agree, I don't see what rounded pfps has to do with collapsible comments, it seems like it was inadvertently left in.

Issue #133, opened 2023-07-17
[MOD] Notifications panel: fetch and load content using async


aclist:
Current panel implementation uses an iframe to force the entire notifications page of the site into the view.

Proposal:

Fetch the names/messages elements of that page in async, use a progress animation in the panel itself, then load in the contents when ready. Also gives more fine-grained control over how the elements are arranged and manipulated, as they could just be stored in arrays rather than rendering the whole page.

aclist:
@artillect @minnieoo End-to-end prototype is ready in this branch: https://github.com/aclist/kbin-kes/tree/patch-notipanel

Content is fetched async and stored in arrays. A spinner animates while loading. When ready, the content loads in.

The code is a lot more compact now and takes a more function => callback oriented structure.

Now we need help from the CSS masters: currently no styling is being applied to the messages in the panel. I think a grid template would probably do the trick. We can lay them out in the same way as the old CSS and give them the same class names if necessary--so that it looks identical, but we're actually loading the content from arrays instead of an iframe.

The "Read" and "Purge" buttons will be easy to add, they are just a POST request to the server.

aclist:
Also, the spinner animation could be anything, I just threw that one in as an example. Could center it in the panel or do whatever with it, too.

aclist:
- [x] Add time element after name
- [x] Add unique classname to private message
- [ ] Add read/unread classnames -- pending

Completed and will commit those updates later after testing it.

honkaiimpact3rd:
I doubt pagination would work well in a panel from a user perspective, my idea so far is to have infinite scroll or a "show more" button at the bottom that loads the next page of notifications (make it a preference or make it depend on the "infinite scroll" setting in kbin settings? maybe?)

I never had multiple pages of notifications (I purge them regularly) so I have no clue how it really looks like though lol

aclist:
> Infinite scroll

Hmm, that's not a bad idea. I think you could safely fit a small arrow-left/arrow-right icon at the top or bottom too if necessary.

I never had thousands of notifications, either, so don't have any real-world data on that, but the first page tends to hold quite a lot of notifications before you need to reach page 2.

aclist:
Took a look at the unread/read, unread messages just come through with a reduced opacity class:

Read: `section section--small thread`
Unread: `section section--small thread opacity-50`
```
.opacity-50 {
    opacity: 0.5
}
```

So we can just carry that classname through with no change and should be good!

aclist:
[...]

PR #158, opened 2023-07-19
[MOD] Notifications panel with async

honkaiimpact3rd:
Here's how it should look like right now
image

Because of Kbin's styling using `!important` (which I explained in #133) it doesn't work properly
image

Removing it from `padding` and `display` fixes it
image
image

### DIFF
--- a/mods/notifications-panel.user.js
+++ b/mods/notifications-panel.user.js
@@ -151,9 +151,7 @@ async function insertMsgs (response) {
     let iff = document.querySelector('.notifications-iframe');
     let parser = new DOMParser();
     let notificationsXML = parser.parseFromString(response.responseText, "text/html");
-    console.log(notificationsXML)
     let currentPage = notificationsXML.all[6].content.split('=')[1]
-    console.log(currentPage)
     let currentPageInt = parseInt(currentPage)
     let sects = notificationsXML.querySelectorAll('.notification');
     if (sects.length === 0) {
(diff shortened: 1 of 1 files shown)


aclist:
I've updated it to check the number of pages and handle these cases:

1. Only one page of notifications => draw no arrows
2. More than one page of notifications => draw forward arrow
3. If current page != 1  => draw back arrow
4. If no notifications => print generic "No new notifications!" message

Number 4 is styled with the selector `.noti-no-messages`. I didn't add any alignment or anything to it, so you'll have to choose where to display this or whether to center it in the panel.

I haven't added fetching the pages and reloading the content yet, but this should get the basic interface in place.

honkaiimpact3rd:
Here's what I did in for notification context, right now it only shows snippets of messages but not if it's a message, reply, or new comment so it's a bit messy

Option 1 - Icons
image

Option 2 - Text
image

Option 3 - Icons with text
image

Not shown on screenshots, but with option 2&3 we would make the panel wider so it handles long usernames better

aclist:
The text "mentioned you..." seems a bit verbose, I know that's the kbin default, but can we improve on that? Option 1 looks pretty good and highly readable if we can get it all aligned properly.

Hmm, I guess of the panel should definitely expand to fit arbitrary text dimensions, not sure. You would know better than me the implications of this.

Looks like the noti-header is not filling the whole frame, either.

aclist:
What about putting the icon next to the message body? Would that make more sense, since it denotes the type of message, whereas the sender name is just its own thing?

Also, what about an arrow pointing down and to the right, i.e., inward towards the message snippet? The arrow pointing out and to the left kind of looks like it's a "reply out" ( == reply back) versus "incoming reply".

honkaiimpact3rd:
>Looks like the noti-header is not filling the whole frame, either

noti-header is filling, but it looks like it's not because of the scrollbar, right now it's siblings with notifications so you would need to put all notifications together like this and make that scrollable instead of the entire panel

image
image

>what about an arrow pointing down and to the right, i.e., inward towards the message snippet? The arrow pointing out and to the left kind of looks like it's a "reply out" ( == reply back) versus "incoming reply".

I couldn't find an icon that would work for that that so what if the icon was just flipped? Akkoma (if you don't know what it is, it's a fedi microblogging platform like Mastodon) uses exactly this.

> What about putting the icon next to the message body? Would that make more sense, since it denotes the type of message, whereas the sender name is just its own thing?

Something like this? I guess it could work, but the circle design could be changed a bit

image

aclist:
> noti-header is filling, but it looks like it's not because of the scrollbar, right now it's siblings with notifications so you would need to put all notifications together like this and make that scrollable instead of the entire panel

Makes sense, noted

> I couldn't find an icon in font awesome that would work for that that so what if the icon was just flipped? Akkoma (if you don't know what it is, it's a fedi microblogging platform like Mastodon) uses exactly this.

Seems reasonable to me 

> Something like this? I guess it could work, but the circle design could be changed a bit

Seems good, maybe the circle is not needed?

aclist:
I hooked up the forward/back buttons to query the right page and load in the content. I think this just leaves the read/purge actions. I'll also take a look at rearranging the div structure later so that the scrollbar is below the header. I added something called `noti-panel-holder` which holds everything inside the panel, letting us remove this holder when loading in new pages and repopulate it.

aclist:
Alright, I think I added everything that was missing and reworked a few things so that animations are always smooth when changing pages, reading/purging content, etc. Also added the API calls for reading and purging, and various error handling for if there are no unread messages, and fixed some click event bugs so that the panel doesn't try to reload itself when clicked. 

This should be everything functionality-wise, besides the message type icons and any remaining styling.

PR #163, opened 2023-07-19
[MOD] Mobile cleanup

twelph:
Hiding two of my least used buttons to help declutter on mobile.


aclist:
Glad you got set up and working! Will review and make some suggestions later on.

aclist:
Nice mod. Sorry for the wait. Had to test it and make the necessary changes to fit the framework. Explanation follows:

Take a look through the added/changed files I pushed in to your fork to get a better idea. First, since these mods are closely related to one another, it makes sense to consolidate them into one and let them inhabit the same menu with two checkboxes (one for each icon to toggle). 

Secondly, you need to add the mod metadata to `manifest.json`, which does the main business of setting up your mod's area in KES. Here's what it looks like:

```
  {
    "name": "Mobile view cleanup",
    "author": "shazbot",
    "version": "0.1.0",
    "label": "Mobile view cleanup",
    "login": false,
    "recurs": false,
    "desc": "Hide elements to reduce clutter in mobile view",
    "namespace": "mobilehide",
    "entrypoint": "mobileHideInit",
    "page": "debug",
    "fields": [
      {
        "type": "checkbox",
        "key": "filter",
        "checkbox_label": "Filter icon",
        "label": "Hides filter icon in navbar"
      },
      {
        "type": "checkbox",
        "key": "view",
        "checkbox_label": "View icon",
        "label": "Hides view icon in navbar"
      }
    ]
  }
```

`namespace` is some unique name where the mod's settings are stored.
`entrypoint` is the name of the toggle function by which we enter your script. Further explanation below.
`fields` are the fields we want to populate. I suggest consolidating the options together into one "mod" here, since they both related to hiding elements in the same area. As you can see, there are two checkboxes.
Each `field` must be given a unique `key`. This key stores the atomic settings for that option underneath the mod's namespace, so `mobilehide.filter`holds the state of the filter checkbox, and `mobilehide.view` holds the state of the view checkbox.

`label` and `checkbox_label`:

The first defines a label above an option, while the second is adjacent to a checkbox and gives it some other descriptive name.

`recurs`: only applicable for things like comments or posts, which update or appear in the event of infinite scroll, so settings may need to be applied again. Doesn't apply to static elements like the navbar icons, so we use `false` here.

We always need an entrypoint function in every mod. In this case, `mobileHideInit`. This corresponds to the manifest above.

When the mod is off:

1. state is `false`
2. Every time the page reloads, we enter via `mobileHideInit`
3. `toggle` is false => branch selects `teardown` function
4. teardown function necessarily sets both elements to `block`, i.e., visible. When the mod is off, these must always be set to visible.

When the mod is on:
1. state is `true`
2. Every time the page reloads, we enter via `mobileHideInit`
3. `toggle` is true => branch selects `setup` function
4. If a checkbox is on, set the style to `none`, i.e., hidden

When the mod is already on and a checkbox is ticked:
1. Every time a checkbox is toggled WHILE the mod is on, we step through the `setup` function again to dynamically update content on the page. 
2. If the checkbox is now off, i.e., `false`, the style is set to `block`. This extra logic is necessary to support dynamically update settings while a mod is still on, without having to switch it on and off again.

`getModSettings()` is a utility function provided by KES to let you fetch your mod's settings from the shared storage. You give it the namespace you set above as an argument, in this case, `mobilehide`.
Then you filter out the `keys` you set in the manifest to get the value of that setting. For a checkbox, it's a boolean (true/false). For other input field types, it might be a number, color hex value, text string, etc.

So with `const settings = getModSettings('mobilehide')`, we are assigning the contents returned by getModSettings for the `mobilehide` namespace to the constant variable `settings`, then we can use the contents of this variable later to read out our mod's settings.

So, in summary:

- Always have an entrypoint (toggle) function
- The name of the entrypoint function should be the same as the one you set in the manifest
- Use the namespace/key parameters to state where to store your mod's custom settings
- Poll the contents of settings using `getModSettings()` and use the value in a conditional statement to perform some setting

As for the aria-label: you are just using it as a selector to find the button. You aren't hiding the label per se, you are hiding the whole button, so insofar as you are hiding the button, this element doesn't exist onscreen to be read by a screen reader, so there is no issue here. (If I understand your question correctly.)

I went into detail here to give you an end-to-end idea of the basic workflow. It may seem a bit daunting at first, but it's pretty formulaic and templated once you get the hang of it. The manifest is largely the same every time, and the mods always get well-defined functions which enter, setup, and tear down for the most part.

Hope this is clear!

aclist:
Any others you wish to include in this current version?

aclist:
You did great. Actually, I forgot to mention, you used the ternary operator! You sure you're a beginner? :laughing: 

In principle, yes, the toggle function is similar, but the hide thumbnails mod only does one thing: hide or show, no detailed settings. As soon as you add a custom setting that changes something other than do/don't do, you have to add extra control flow because not only are we going to step through the mod each time, but we also have to apply each of its settings. Perhaps I overcomplicated it by combining these into one, but I think it is the way to go here.

The main thing that was missing was the entry in the manifest, which controls everything. Without the manifest metadata, a script won't even be added to the menu.

aclist:
It looks pretty good so far. The nice thing about this system is that it is modular, so you can slot in more features on a rolling basis pretty well. I'm just here to massage the results into a cohesive format so that mods play together nicely, but the ideas and design are yours.

aclist:
I look forward to seeing what you and your AI buddy come up with!

PR #184, opened 2023-07-23
[MOD] Move magazine federation warning to sidebar

jansteffen:
## What's added/changed

When viewing a magazine from another instance, kbin displays a warning at the top of the feed. This mod moves the warning to the sidebar's magazine info panel.

demo

## To-do list/next steps

For now I've kept this script as simple as possible, but I've thought of a few ways this could be expanded upon further:

- [ ] Add option to hide the warning completely
- [ ] Add the "View on original instance" link as a button next to the subscribe/block buttons
- [ ] Move/hide federation warning when viewing a user profile (currently unaffected)

## Checklist

Add-ons (mods):
- [x] `manifest.json` was updated to include the add-on's metadata

I'm still a noob at this so please bear with me if anything is wrong 😅


aclist:
This is great, thank you so much for your submission! Works well and looks clean. We can definitely include this in the upcoming 2.1.0 release.

One small thing I would suggest adding to make this more robust is check the window location at the top of the initial function, because this warning box can only appear on magazines, correct? So the subdomain must necessarily contain the substring `m`:

If `window.location.href.split('/')[3]` is not `m`, simply `return` without doing extra logic.

This is not really important on a small script, but could be good practice going forward on larger scripts in order to save cycles, since you can short-circuit the entire business logic if this condition is false.

> Add option to hide the warning completely

I think this would be great to add if you have time. We are moving towards a more accessibility-oriented "let the user decide" approach with the next release, rather than "this is what we like" approach with mods that only have one setting. You'll need to use `getModSettings`, set up a namespace key in the manifest, and probably use two radio buttons. Let me know if you need further help with that.

jansteffen:
I'm having trouble getting my mod to actually show up in the KES menu with my local version to test radio boxes. Your repo says the `kes.user.js` file is autogenerated, is there a build script I need to run?

aclist:
Yes, `gen.sh` (bash shell script) is used to generate it, but it's intended more as a pre-release script for the maintainer. By default it interpolates the URLs with the name of the current branch so that it can fetch the remote resources from there.

I would just run that so that it fills in the function names and everything else, then replace the URLs to the manifest and to your script:

```
Before:
// @require      https://github.com/aclist/kbin-kes/raw/testing/mods/move-federation-warning.user.js
After:
// @require      https://github.com/jansteffen/kbin-kes/raw/testing/mods/move-federation-warning.user.js

Before:
const manifest = branchPath + helpersPath + "manifest.json"
After:
const manifest = https://raw.githubusercontent.com/jansteffen/kbin-kes/testing/helpers/manifest.json
```

Everything else will be loaded from this repository.

In general, probably the fastest way of testing locally is to replace URLs with your testing versions. You raise a good point, so I'll expand the build script so that it replaces not just the branch name, but the whole repository name, so that you could just update your own repository and run `gen.sh` and have a correct `kes.user.js` that points to all of your own resources.

aclist:
Alright, I've updated the `gen.sh` in the `testing` branch to interpolate your own repository into the URLs, so it should generate all the right resources now and you can just load the new `kes.user.js` in your browser.

jansteffen:
Running the `gen.sh` script is throwing an error for me: 

    gen.sh: line 162: jq: command not found

aclist:
Sorry, I didn't add any dep checks. It requires `jq` to parse the manifest. I could drop this dependency and parse it natively if this package is hard to install, but it should be widely available.

aclist:
Alright, pushed a new version. I resorted to parsing the manifest natively using awk, so the dependency should clear.

jansteffen:
Allright thanks, everything seems to work correctly now on my end, feel free to test it again now.

aclist:
Works great. Thanks for your patience. Approving.

PR #191, opened 2023-07-25
[MOD] Hover Indicator

minnieoo:
The hover indicator mod adds an outline to almost every element when it is hovered on. This can help with accessibility.

Features:
- Able to change outline color
- Able to change outline thickness

image
image


aclist:
By the way, could you check out my little project too?
https://github.com/aclist/kbin-kes/raw/tests/omnibar/tests/omni.user.js

It's a standalone test at the moment, but this will be hooked up to KES logic so that it can be summoned at will rather than on page load.

Use arrow up/down to navigate, or type in the field to filter search results. Hit enter to jump to the desired mag. You can also use the mouse and click. If you hit Tab or click off the list it'll remove it.

Should look good and work well on mobile, too.

There are a few cosmetic bugs with mouse hover, but otherwise should be working. One thing to test is how it looks on other themes :roll_eyes:  (Looks like garbage on the purple theme, I know.) Hopefully you can help with that...

minnieoo:
I just tested it, appears to be working good. What is it's purpose?

aclist:
It's called an omnibar. Fast jump to one of your subscribed mags.

1. Tap some key to summon omnibar
2. Find your mag
3. Fast jump to it

Maybe that wasn't obvious enough?

minnieoo:
o yea i see that now. i didn't realize bc it doesnt yet display all the subscriptions. looks like it is working well tho

aclist:
> i didn't realize bc it doesnt yet display all the subscriptions

Hmm, can you elaborate? I only have a few subscriptions tested. If you have a ton, it should be scrollable. But the point is you type to filter down the list. 

I'm wondering if you're describing a bug or something else.

minnieoo:
oh, yea i do have quite a bit of subscriptions, upon reloading a few times I can scroll it as it shows more mags. it shows mags i am not subscribed to, and my actual subscriptions aren't in there as can be seen when i type to find a few. also, how do i spawn this without reloading page, if i can atm?


https://github.com/aclist/kbin-kes/assets/137783738/360d41d9-f837-4fdf-b98c-0656c88ed6c6

aclist:
Oh hahaha, I am so dumb, I hardcoded it to my subscriptions list and was just randomly adding and deleting magazines lol. Idiot alert...

OK, so I need to check the user's actual name and handle the scrolling better.

As for opening it on demand, I didn't add that yet, will probably be the last thing I do. Just a test now.

aclist:
Alright, I've fixed it a bit (v 0.2) thanks to your astute observations lol. It should fetch your actual subs list now, be more compact, and have a scrollbar. It won't load immediately on the page. Hit `Tab` (using that as a test key for the time being) to summon it, then `Tab` again to close it. It takes a second right now because it fetches the list each time, but will change it to cache the results to make loading faster.

minnieoo:
working good now, all my subs are there! tab working as well

aclist:
Thanks! I'll just add a few more things and eventually integrate it into KES. Does `Tab` seem problematic? We can basically hijack any key, but never sure what kinds of keys people expect to have access to for other purposes. 

Is this hover indicator mod ready to go, or are you still working?

minnieoo:
hover indicator is ready, i worked out the conflict. now i am onto another accessibility script :D one for adjusting the sites color, warmth, hue, contrast, saturation, etc etc

i dont ever use my keyboard for navigation, so i am honestly unsure if it would be an issue for others. for me, it isn't an issue. i would be using tab for text areas if i could, but kbin doesn't support that for some reason, so it's no issue. i don't use it. perhaps the user should be able to set their own shortcut?

aclist:
> hover indicator is ready, i worked out the conflict. 

OK, cool, will review and merge. Released 2.1.0 earlier today, so this will go into the next release.

> now i am onto another accessibility script :D one for adjusting the sites color, warmth, hue, contrast, saturation, etc etc

Sounds good!

>as for the tab key, i am honestly unsure if it would be an issue for others. for me, it isn't an issue. i would be using tab for text areas if i could, but kbin doesn't support that for some reason, so it's no issue. i don't use it. perhaps the user should be able to set their own shortcut?

Yeah, I just thought of that. We should be leaning towards letting you choose what you want rather than enforcing anything. I'll just offer an assortment of meta key options and let them choose, I think. I don't want to set up a whole "input any key combo" feature, though, as it'll introduce problems. You definitely raise  a good point that tab should be used for navigation.

You stealth edited that post like 5 times, I saw it change before my eyes :P

You're the exact opposite of me, I don't use the mouse for anything.

minnieoo:
>You stealth edited that post like 5 times, I saw it change before my eyes :P

LMFAO you saw that 😭 yea i realized too late i didn't word what i meant right several times in a row apparently hahahah

>I don't use the mouse for anything.

i use my trackpad on my laptop for everything 😭😭

minnieoo:
@aclist can the range input increment by 0.1? cant seem to get it to work like that

aclist:
Should be, yeah; `<input type="range" min="1" max="100" step="0.1">`. Not working?

aclist:
Looks like step support wasn't added when somebody created the range function, will add it now.

minnieoo:
o okay nice :D ty ty

helpers/manifest.json
```diff
@@ -924,5 +924,34 @@
...
+    "login": false,
+    "recurs": false,
+    "desc": "Apply an outline to elements that are being hovered over.",
+    "namespace": "hover",
+    "entrypoint": "hoverIndicator",
+    "page": "accessibility",
+    "fields": [
+      {
+        "type": "color",
+        "initial": "#FFFFFF",
```
aclist, line 942: Same question as the other script: can we start this at the default hover color (e.g. `--kbin-vote-text-hover-color`)? Feels like `FFFFFF` might not render on the white theme.

aclist (changes requested):

aclist:
Haven't tested this yet but will do so today.

aclist:
Just merged this so I could test it a bit more quickly, everything seems essentially fine, but did you check out some of the links on the navbar (e.g. KES icon, notipanel bell) or below post comments (e.g. reply, boost, more)? When there are links within outer link elements, it creates a kind of nested box. Notably on the KES icon, which is an outer element, an inner holder element, then a wrench icon, then the "updates available" notifier, so it's a total of four boxes:

a

Any thoughts on that? This also seems to happen on upvote/downvote buttons, though to a lesser degree, because the "number" count and the "arrow" icon are sub-elements of the outer vote box, even though only the vote box registers as a click here. It seems like ignoring icons (`<i>`) might be worthwhile?

It's also somewhat noticeable on avatar icons in thread replies.

aclist:
By the way, the manifest generator should have all basic JSON functionality for generating the different field types now. The visual preview isn't working yet, but the content copied to clipboard should be correct and ready to drop in to any manifest.

minnieoo:
> but did you check out some of the links on the navbar (e.g. KES icon, notipanel bell) or below post comments (e.g. reply, boost, more)? 

i noticed some nested boxes but it didnt seem to look that bad to me.
firefox_oNwEu3Jf6T
firefox_rMZBjFlTvv

maybe it could be that it only puts the outline on the outermost parent? not sure how well that would work. not sure how easy or hard it would be to select the correct parent of something since everything is nested.

>It seems like ignoring icons (<i>) might be worthwhile?

yea thats fine. i added it for cases such as this, and really any time that you are needing to click anything represented as an icon:
firefox_5H5kgP3CuX

>the manifest generator should have all basic JSON functionality for generating the different field types now

awesome

aclist:
Yeah, that is a good point. You don't want to ban all icons altogether. It doesn't seem like an issue on too many elements, mostly things where there is a span tag wrapping a link or an icon, so maybe just adding specific exclusions for those would be better than trying to have some abstracted logic that might potentially fail on other stuff?

For example: 
- if navbar, only hover the outer elements
- if thread comment buttons, same thing?
- if vote buttons, only the outer container?

On post headers, the header is actually a header box, plus the thread link itself, and optionally the metadata showing that URL it originated from, and you would potentially want to click either the thread link or the metadata, but is the outer hover on the h2 element needed? Hope this description makes sense.

If I understood correctly, the hovering should aid targeting clickable elements, right? If that's the case, it seems reasonable to reduce it to the "minimum clickable element" rather than all elements. And if the hover is meant to just highlight the focused element for tracking purposes, it still seems like too much nesting might actually make it hard to see what you are hovered on.

aclist:
One other area to think about:

If you have a list of items with links within them, the links are the only clickable part:

```
<ul>
    <li>
        <a>
</ul>
```

For example, in the kbin footer, if you highlight the `li`, it's not actually a clickable link until you hover the `a` itself. So should the `li` actually be highlighted here, or only the inner `a`?

It depends on whether the answer is "hover for clickability" or "hover for visibility." Seems like it might be one or the other, or both, depending on certain situations.

minnieoo:
>If I understood correctly, the hovering should aid targeting clickable elements, right?

actually i didnt make it with that in mind, i just made it with the idea that no matter what you're hovering over, it will indicate the element you're on (thats why you can hover over the comments and more), but it would make more sense to do it for just clickable things

aclist:
Not necessarily, maybe we can find more accessibility information about these kinds of needs. I just assumed that targeting something means you want to click on it, but maybe it's more of a navigational aid? By the same token, I'm not sure when would be a situation where you need to know that you are hovering an icon within a button, but maybe there is something to this if we think about it?

minnieoo:
> navigational aid

thats the purpose i intended it to have yea. 

>I'm not sure when would be a situation where you need to know that you are hovering an icon within a button

i agree i dont think this is necessary, i just didnt get super complex with it and rule out those situations when i should have, and for some things i didnt know how

aclist:
That is fine, I don't want to change it. I will just tweak some of the more heavily nested areas, if that is OK?

minnieoo:
oh im just saying what i originally was thinking. i dont care at all what is done to it, do whatever

aclist:
OK, cool. I didn't see your message until now for some reason.

aclist:
OK, I added a few exclusions to the hover conditions on testing if you wanna check it out.

minnieoo:
looking good

Issue #203, opened 2023-07-30
[CHORE] Preview images


aclist:
@minnieoo 

yea sure i can do that. do we want them in `/images` or perhaps a separate folder like `/screenshots` or `/previews`? there might be quite a few of them, not sure if it matters if they are mixed up with the current example images and the kibby images but thought i'd ask anyway. also what is the field called? links? would i be doing something like `"link": "/images/mod-preview.png"` on the mods i take the sc's for?

_Originally posted by @minnieoo in https://github.com/aclist/kbin-kes/issues/194#issuecomment-1657036710_

minnieoo:
should /previews be within /images?

aclist:
Yeah, definitely. `/images/previews/XXX`

aclist:
Duplicate of 204

PR #204, opened 2023-07-30
Add preview images

minnieoo:
Adds preview images to most mods for clarity

solves #203 "[CHORE] Preview images"

image


aclist:
My instructions were a bit unclear on this one. When I said "try to keep them small," I was talking about the compression, not the crop tightness. Some of these are cropped too tightly, so there is no context around them to see what the change is relative to the rest of the page. For example, if you just see a picture of a vote button or a picture of the kibby icon, it doesn't give enough surrounding context for how that works on the page.

This is a common mistake due to being "too close to the code." We are intimately familiar with these options and what particular elements on the page do, but some users might have no clue where the pictured element is supposed to fit relative to the rest of the page.

I also think we haven't done our job well if we have to redact/paint over images to hide information. It indicates that we need to stage the images better, be it through dummy accounts/posts, to make the context look clear, rather than drawing lines over images, which doesn't give a good look.

I'll reshoot some of these and update. Manifest itself and structure looks great.

minnieoo:
alrighty sounds good

aclist:
Thanks, sorry to be a stickler and give negative criticism :worried:

PR #226, opened 2023-12-04
[Mod] Fix code blocks

Pamasich:
## What's added

Added a mod that fixes a long-running issue with code blocks federated from Lemmy on /kbin. 

Lemmy includes extra `<span style="color:#323232;">` tags enclosing every line in their federated code blocks which /kbin just displays in plaintext. See [this comment](https://kbin.social/m/programming@programming.dev/t/623282/Implementing-Tic-Tac-Toe-with-170mb-of-HTML-no-JS#entry-comment-3548488) for an example.

This mod fixes that by removing those tags.

## Checklist

- [x] ESLint did not produce any errors
- [ ] `manifest.json` was updated to include the add-on's metadata

Didn't update manifest.json since the documentation says to only submit the mod folder and the manifest's contents seem like the mod's own json file is just copy-pasted there. If I need to do something in this file after all, please tell me, I'll follow the documentation for now and leave it alone.

### DIFF
--- a/mods/fix-codeblocks/fix-codeblocks.user.js
+++ b/mods/fix-codeblocks/fix-codeblocks.user.js
@@ -1,4 +1,4 @@
-function fixLemmyCodeblocks(toggle) {
+function fixLemmyCodeblocks (toggle) {
     // included everything inside here because of the private functions advisory
 
     const testPattern = /\n?<span style="color:#323232;">(.+\n)+<\/span>\n?/;


Pamasich:
Ah, I used the current main branch as the base for the code, but the documentation says to do a pull request to the testing branch which is behind the main one. Didn't expect that to also include the commits made to the main branch since.

This is my first pull request, so not sure how this stuff works, will this resolve itself once someone merges main into testing?

Pamasich:
I'll just copy my changes into a new branch based off the testing branch tomorrow and submit a new pull request, will also let me pseudo-squash the commits.

aclist:
> Ah, I used the current main branch as the base for the code, but the documentation says to do a pull request to the testing branch which is behind the main one. Didn't expect that to also include the commits made to the main branch since.

Generally speaking, testing should be ahead of main and release candidates are forked off of testing and submitted into main as PRs, but testing might also include some experimental features that are of use to developers or which have no purpose being in main. Seldomly, PRs containing patch hotfixes are submitted against main to fix the live version if something breaks immediately upon release, and these later need to be backported into testing, which might lag behind at times.

Documentation could stand to be clarified here. I think there is also some drift between static documentation contained in main and testing, which should be normalized.

> This is my first pull request, so not sure how this stuff works, will this resolve itself once someone merges main into testing?

Yes

PR #227, opened 2023-12-05
[Mod] Fix code blocks

Pamasich:
## What's added

Added a mod that fixes a long-running issue with code blocks federated from Lemmy on /kbin. 

Lemmy includes extra `<span style="color:#323232;">` tags enclosing every line in their federated code blocks which /kbin just displays in plaintext. See [this comment](https://kbin.social/m/programming@programming.dev/t/623282/Implementing-Tic-Tac-Toe-with-170mb-of-HTML-no-JS#entry-comment-3548488) for an example.

image

This mod fixes that by removing those tags:

image

## Checklist

- [ ] `manifest.json` was updated to include the add-on's metadata

I'm a bit confused by manifest.json being in this mod-specific checklist, you might want to edit either the checklist or the documentation. The latter claims manifest.json is automated and I should only be submitting the mod's folder and no other files, so I'm leaving manifest.json alone unless you confirm I should change something about it.

### DIFF
--- a/mods/fix-codeblocks/fix-codeblocks.json
+++ b/mods/fix-codeblocks/fix-codeblocks.json
@@ -7,5 +7,5 @@
   "login": false,
   "recurs": true,
   "entrypoint": "fixLemmyCodeblocks",
-  "page": "General"
+  "page": "general"
 }
--- a/mods/fix-codeblocks/fix-codeblocks.json
+++ b/mods/fix-codeblocks/fix-codeblocks.json
@@ -1,0 +1,11 @@
+{
+  "name": "Fix Lemmy Code Blocks",
+  "author": "Pamasich",
+  "version": "1.0.0",
+  "label": "Repair Lemmy code blocks",
+  "desc": "Lemmy federates special <span> HTML tags on each line of its code blocks which /kbin displays in plaintext, resulting in code blocks being hard to read. This mod removes those tags.",
+  "login": false,
+  "recurs": true,
+  "entrypoint": "fixLemmyCodeblocks",
+  "page": "General"
+}
--- a/mods/fix-codeblocks/fix-codeblocks.user.js
+++ b/mods/fix-codeblocks/fix-codeblocks.user.js
@@ -1,0 +1,37 @@
+function fixLemmyCodeblocks (toggle) {
+    // included everything inside here because of the private functions advisory
+
+    const testPattern = /\n?<span style="color:#323232;">(.+\n)+<\/span>\n?/;
+    const startTagPattern = /^\n?<span style="color:#323232;">/;
+    const endTagPattern = /\n<\/span>\n?$/;
+    const combinedPattern = /^<\/span><span style="color:#323232;">/gm;
+    // matches any line (or rather the start thereof) except the first one
+    const startOfNewLinePattern = /(?<=\n)^/gm;
+
+    const attribute = "data-code-fixed";
+    const startTag = '<span style="color:#323232;">';
+    const endTag = "</span>";
+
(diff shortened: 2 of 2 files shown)


aclist:
Thank you, this seems great.

> I'm a bit confused by manifest.json being in this mod-specific checklist, you might want to edit either the checklist or the documentation. The latter claims manifest.json is automated and I should only be submitting the mod's folder and no other files, so I'm leaving manifest.json alone unless you confirm I should change something about it.

The PR submission checklist was not updated after the recent change to the manifest system. You did it correctly, but the checklist needs to be updated to reflect this, as you noted.

aclist:
@Pamasich 

Per https://github.com/aclist/kbin-kes/issues/237, it seems the current regex is brittle. It is not always the case that:

- The hex color for the span tag will be `323232`
- A single span tag will appear per line
- The closing span tag will start on a newline

Given that the mod also inserts span tags back into text when it is off, the "original" code block listed in the report is erroneous, so to see the actual format, we have to open the page on a browser with no KES installed. It looks like this, inclusive of leading newline:

```

<span style="color:#0086b3;">Vec</span><span style="color:#323232;">::new();
</span>
```

As you can see, not only is the hex color inconsistent, but the first line following the newline contains multiple span tags in succession.

Whereas the second block conforms to the old "standard":

```

<span style="color:#323232;">Vec.new()
</span>
```

A blunt reconfiguration of the existing logic yields something like:


```
const testPattern = /\n?<span style="color:#[a-zA-Z0-9]{6};">/;
const startTagPattern = /^\n?<span style="color:#[a-zA-Z0-9]{6};">/;
const endTagPattern = /\n<\/span>\n?$/;
const combinedPattern = /\n<\/span><span style="color:#[a-zA-Z0-9]{6};">/gm;
const combinedPattern2 = /<\/span><span style="color:#[a-zA-Z0-9]{6};">/gm;
---snip---
    code.innerText = code.innerText
                .replace(startTagPattern, "")
                .replaceAll(combinedPattern, "")
               .replaceAll(combinedPattern2, "")
                .replace(endTagPattern, "");
```

While capturing 6 alphanumeric chars is sufficient to cover any hex color code, the problem becomes increasingly complex if we consider that a string literal span tag may be legitimate code inside a pre block, i.e., if the author's code itself consists of raw HTML and span tags. Presumably this would look something like `<span...><span...>...</span></span>...`, and you would have to use multiple regex capturing groups to capture the outer enclosing span tags (those inserted by kbin) for each occurrence.

If the hex code is dynamic, this also complicates re-insertion of span tags after changing the innerText, unless the entire code block is stored verbatim somewhere. As for this issue, I think it is not as intractable as it seems, as you could perhaps, instead of overwriting the innerText, simply overlay the new code block on top of the old one and block the visibility of the old one via CSS filters, etc. Then teardown would consist of removing the overlaid code block without having to destructively re-insert text.

In any event, the cosmetic issue can be relatively easily fixed if you are OK with an equally brittle (but more cosmetically satisfying) regex that might destroy legitimate span tags in HTML code examples.

Another alternative would be to stop replacing the text programmatically altogether, and use an XML request to fetch the actual contents of the post body from the off-site instance. Obviously, this adds significant load time, particularly if it has to happen for multiple code blocks across a thread. However, perhaps the span tag sanitization could be user-invoked instead of dynamic, so that the XML request is performed on-demand when the user clicks a "sanitize code block" button or the like.

Pamasich:
Weird, Github didn't notify me of this comment at all. Just checked the latest issues and saw the one about the mod. So, unless I figure out why I didn't get notified, replying here might not be the best way to contact me.

I'll look into this issue over the next week.

Issue #237, opened 2023-12-27
[BUG] "Repair Lemmy code blocks" fails to remove some <span> tags


pbackus:
**Describe the bug**

The "Repair Lemmy code blocks" mod should remove all of the `<span>` tags inserted into code blocks federated from Lemmy instances. However, some `<span>` tags are not removed.

**To Reproduce**

1. Go to <https://kbin.social/m/programming@programming.dev/t/726929/The-C-scope-resolution-operator-is-beautiful>
2. Enable "Repair Lemmy code blocks" in KES.
3. See that there are still `<span>` tags inserted by Lemmy in the displayed code block.

Without "Repair Lemmy code blocks" enabled, the first code block in the linked post looks like this:

```
<span style="color:#323232;">
</span><span style="color:#323232;"><span style="color:#0086b3;">Vec</span><span style="color:#323232;">::new();
</span>
```

With "Repair Lemmy code blocks" enabled, the first code block in the linked post looks like this:

```
<span style="color:#0086b3;">Vec</span><span style="color:#323232;">::new();
```

**Expected behavior**

No `<span>` tags should appear in the linked post's code blocks when "Repair Lemmy code blocks" is enabled.

**KES version:**
- 3.1.0

**Extension and version:**
- Extension: TamperMonkey
- Version number: 5.0.0

**Desktop (please complete the following information):**
 - OS: Debian Linux 12.4
 - Browser: Firefox
 - Browser version: 122.0b3

aclist:
Thank you for the report. I have brought this to the attention of the mod's author and suggested a few possible workarounds.

Pamasich:
I'll look into this over the next week.

This does look like it complicates things quite a bit.

aclist:
Resolved in 3.2.0

PR #238, opened 2024-01-03
(fix-codeblocks) Fixes in response to #237

Pamasich:
[example]: https://kbin.social/m/programming@programming.dev/t/726929/The-C-scope-resolution-operator-is-beautiful

This pull request resolves the issues seen in #237.

With these changes, the [example] brought up there works perfectly.

Specifically, the two issues that were made apparent in the [example] are
1. the mod determines erroneous tags by a specific color, but they can have any color.
2. the mod assumes there's one erroneous closing and one ending tag per line only, but there can be any number of them.

To make sure these can be restored correctly on teardown, the mod now preserves the original code and merely hides it, instead adding the fixed code to a new element. On teardown the new element is then simply deleted and the original one made visible again.

I've also fixed an issue with the overall test the mod makes first to determine if a code block needs to be fixed. It's supposed to match the entire code, but was returning true if only a part of it matched. Which is why the example added tags that never existed on teardown for OP, it should have never recognized that code block in the first place if it worked correctly.

### DIFF
--- a/mods/fix-codeblocks/fix-codeblocks.json
+++ b/mods/fix-codeblocks/fix-codeblocks.json
@@ -1,7 +1,7 @@
 {
   "name": "Fix Lemmy Code Blocks",
   "author": "Pamasich",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "label": "Repair Lemmy code blocks",
   "desc": "Lemmy federates special <span> HTML tags on each line of its code blocks which /kbin displays in plaintext, resulting in code blocks being hard to read. This mod removes those tags.",
   "login": false,
--- a/mods/fix-codeblocks/fix-codeblocks.user.js
+++ b/mods/fix-codeblocks/fix-codeblocks.user.js
@@ -1,8 +1,10 @@
 function fixLemmyCodeblocks (toggle) {
-    const testPattern = /^\n?<span style="color:#[0-9a-fA-F]{6};">(.+\n)+<\/span>\n?$/;
-    const startTagPattern = /^\n?<span style="color:#[0-9a-fA-F]{6};">/;
-    const endTagPattern = /\n<\/span>\n?$/;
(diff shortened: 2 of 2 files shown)


aclist:
Thank you for working on this despite kbin having sporadic uptime issues over the last week. Looks good over here. I will see to integrating it in the next release over the weekend.

Pamasich:
@aclist I noticed an oversight with how the original block just gets hidden now. If the mod is run multiple times, the hidden code would still get fixed again. Fixed that now in the new commit.

aclist:
Do we have a repeatable way of unit testing different span tag scenarios by simulating posts? It should be enough to create a dummy magazine and account on the other instance, create a post with different varieties of code examples (one per post), and inspect the results on kbin, yes? If you haven't done so already, I think creating a series of tests like this that validate different scenarios (e.g. nested tags) would be a good way of testing the script in the event that future refactoring is needed.

Pamasich:
> Do we have a repeatable way of unit testing different span tag scenarios by simulating posts? It should be enough to create a dummy magazine and account on the other instance, create a thread with different varieties of code examples (one per post), and inspect/programmatically test the results on the kbin side, yes? If you haven't done so already, I think creating a series of tests like this that validate different scenarios (e.g. nested tags) would be a good way of expanding test coverage in the event that future refactoring is needed.

Right, good idea. I'll see to creating some test cases probably next week. I do have a magazine for testing and a user on lemm.ee which is on the latest version of Lemmy and has syntax highlighting turned on.

I'll definitely have to look a bit more into how different things federate too. Whether there's other languages than Rust that have special logic (Java is just the generic one `color:#323232;` per line I had assumed initially, haven't tested more languages than those two yet), and whether any other CSS properties get used.  
I have just tested some Rust example code and found out that I should check for `font-style` too (it's used for comments). So there's definitely some more checking necessary there.

PR #247, opened 2024-01-26
[Mod] Alternative access to the All Content view

Pamasich:
Currently on /kbin
- magazines have three different views people can look at: Threads, Microblog, and All Content.
- you can customize which view is used for the start page, but not for magazines
- the magazine title in the navbar always brings you to the Threads view

I think it makes the most sense for the All Content view to be the default one you get to by clicking the magazine title, while Threads and Microblog would be more like filtered views if I just want to see one of them. Additionally, it would just be nice to be able to have the magazines match the frontpage in what view they default to.

So this mod makes the magazine title in the navbar link to the All Content view instead of the Threads view, and optionally hides the All Content button. The latter is done as it's kind of redundant and just takes up space like this.

This mod doesn't currently affect any links to the magazine in other places than the navbar.

## To-do list/next steps

- [x] Implement mobile support

Btw, do I need to set `recurs` to true for it to work with turbo mode, or is that supported out of the box?

## Checklist

Add-ons (mods):


aclist:
> Btw, do I need to set recurs to true for it to work with turbo mode, or is that supported out of the box?

Turbo mode is enabled by default. The recurs flag only applies for situations where more content is propagated on-demand, such as:

- Infinite scroll (loads new threads/posts)
- Replying to comments (creates a new comment on the page)

As far as I know, those are the main situations where user interaction causes new content to appear on the page without reloading it in place.

Pamasich:
Implemented support for kbin's mobile version, so this should be feature complete now.

Pamasich:
@aclist I'd like to actually test if I did the json correctly, by adding the mod to KES in my browser. So far I just tested this and the other mod through the console (and unit tests in this case). But I can't seem to get KES to load my version of the manifest.json? I haven't really worked with resources in userscripts before, so not sure if I'm forgetting a step.

I have adjusted the resource declaration to this:
`// @resource     kes_json https://github.com/Pamasich/kbin-kes/raw/tests/helpers/manifest.json`

which, if I open it in the browser, shows the correct version containing this mod's data.

However, debugging shows the safeGM getResourceText returns the old version still before my changes, so the mod doesn't show up in KES.

Is this an issue with Tampermonkey or is there something more I need to change besides the resource declaration to make it work?

aclist:
Sorry for the wait; I had to release major version bumps of some other projects during the interim, but have the bandwidth for this now. 

The fix for the initial value being true should be live on testing now if you want to pull, update the manifest to your liking (based on the commit log, assume you want the checkbox to start enabled), and verify before we merge this PR.

Use of classes here is a good and maintainable approach--the main KES logic is itself overdue for an overhaul using an OOP paradigm. What is there now is a concatenation of functions that was bootstrapped by various contributors at the beginning of the project, but will become annoying to maintain should the length of the file grow longer than the ~1K LOC it is now. Currently it is manageable, but at a minimum the logic for creation of modals and other components should be made more generic for recyclability.

Pamasich:
I'll need to fix a bug I encountered before this can be merged.

Pamasich:
Okay, done.

Issue #205, opened 2023-07-30
[BUG] Contents of settings modal not showing


alexander-paterson:
**Describe the bug**
When I click the settings button, I'm not seeing the settings window/modal, I'm seeing markup.

**To Reproduce**
1. Click the settings button.

**Expected behavior**
The settings should open.

**Screenshots**
![image](https://github.com/aclist/kbin-kes/assets/931083/38e08a60-bfc0-4eb9-bcba-28aaa2237cc9)

**KES version:**
- Version number 2.1.4

**Extension and version:**
- Extension: ViolentMonkey
- Version number: v2.15.0

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Firefox
 - Browser version 116.0

aclist:
I cannot reproduce this yet. Have you tried removing KES and reinstalling it from scratch?

aclist:
I took a look at changes in the code between 2.1.3 and 2.1.4, and the only thing that changed was the format of a URL used in two add-ons to make them support domains other than kbin.social, so that change would not have caused the above behavior. My current hypothesis is that the style data became truncated for some reason while downloading the update.

Were there any other errors in the dev console?

aclist:
I was able to get this bug to occur once, but I haven't found steps for reliably reproducing it yet.

aclist:
A few questions:

1. Did you have multiple extensions installed, such as Tampermonkey/Violentmonkey/Greasemonkey at the same time?
2. Did your computer or Firefox crash at any time while it was open and before this error occurred?

alexander-paterson:
1. No, just ViolentMonkey
2. I don't recall anything crashing.

It hasn't happened again after removing and re-adding the script.

aclist:
I have encountered this only once during the intervening three months since you posted it. I confirm that this bug exists, but it is difficult to reliably reproduce, as it seems to be either a race condition or related to throttling from GitHub's API. My surmises are that something fails when pulling in the remote resources, and so the widget inserts the paths to the CSS without actually loading it. I think it might be possible to reproduce this on an initial reinstall, but I have had trouble getting it to happen with 100% reliability. I will continue investigating the root cause.

Pamasich:
I've had this issue too months ago and decided to quickly look into it today.

You can actually reliably reproduce this by changing the "raw/" in `branchPath` to "blob/".

[...]

Issue #132, opened 2023-07-17
Adding to greasyfork


artillect:
Since greasyfork is generally the main place to get userscripts for kbin, it'd be a good idea to get it posted on there. You can set up the greasyfork page to sync with this repo to automatically handle updates. It could also help get a little bit of the load off of GitHub when people install it or update `kes.user.js`, but all of the mods would still have to be loaded from GitHub. 

We could also set up the mods as library scripts on greasyfork, which could open up this up to being more of a standard API for kbin addon packs (and further reduce the load on GitHub as well), but that's a bit of a crazy idea.

aclist:
Interesting, didn't know it autosyncs. The main reason I avoided it was because I figured it would have to be a standalone thing, be separately managed, and doesn't provide proper support channels and stuff to actually keep track of the codebase in a standardized manner. Also got turned off by the ad spam, hahah. I agree that it helps users "find" the tool through their manager and sync.

aclist:
One thing I'd like to pick your brain about is, would this open up the possibility of increased ambiguity with unsanctioned, potentially harmful copycat scripts? What if someone creates a malicious KES clone and posts it on greasyfork under a similar name? What's to stop users from inadvertently finding that one when browsing through their extension search menu? If we could say, "official KES sanctioned by the KES team is only available at this GitHub repository," it might allow us to skirt this issue, since anything on greasyfork or other userscript pages would be de facto not the sanctioned version. 

What I'm getting at here is there is basically no "filter" with userscript pages, it's just a collection of scripts and it's caveat emptor for the user. Whereas it should be a lot harder to make a malicious repository. 

Figured I should elaborate on that here, since that's at least part of what went into my decision to prefer a git repository with standardized entry point page.

I think we could also look at the possibility of checksums. Although not every user is liable to use them, at least we can point to a single point of truth and say, the X.X.X release has the checksum YYYY, anything else is invalid. For starters, I'll go ahead and use the GitHub releases feature to link releases to a specific tag, and list the checksums there.

JasonMaloney:
> What if someone creates a malicious KES clone and posts it on greasyfork under a similar name?

What if you don't post it to Greasy Fork, and someone else does? Better to protect the project name.

aclist:
I'm totally open to whatever you guys wanna do, just throwing out fringe scenarios to do defensive design, as is my wont.

artillect:
[...]

PR #252, opened 2024-03-10
Fix Raw URLs

Pamasich:
As discussed in #205, this commit switches the `/raw/` urls to use `raw.githubusercontent.com` instead. That should fix the problem seen in that issue.

To keep things uniform, I've switched over all the /raw/ urls in the project, not just the ones in kes.user.js.


Pamasich:
@aclist I just noticed while reading the [greasyfork issue](https://github.com/aclist/kbin-kes/issues/132) that you said

> If we reverted to using the github.com TLD instead of raw.githubusercontent.com, this should theoretically work. However, we switched to the second one because the first one is just performing a redirect, so wanted to save a few cycles there. Also, this feels slightly less robust than using the integrity hashes.

Am I interpreting this correctly that KES used to use raw.githubusercontent.com but then switched to /raw/ urls for some reason? Was that just greasyfork, or was there a different reason? Don't want to do a regression here.

aclist:
If memory serves, we flip-flopped back and forth between them a few times because something about the interaction between *monkey's caching and GitHub's deferred caching mechanism was causing some users to see stale versions of files even when an update was pending for KES. I think we reverted to just letting GitHub handle the redirections because it should (?) fetch the more recent file after doing so. 

It still seems like the raw URL should be more consistent and avoid the CSS failing to load, but we should probably test whether an updated KES version triggers the monkey to fetch the new mod versions. It is relatively easy to revert.

Unfortunately, GitHub's caching mechanism is not well documented, so I can't say definitively what is going on.

I belive the @requires will never auto-update on monkey's side (falling back to the locally-stored resources) unless:

1. The entire script is reinstalled
2. The @requires in the updated script are no longer identical to the string literals in the previous version (e.g., if the remote URL to the mod includes a bumped version number or something)
3. The script pulls in a completely new @requires (new mod)

I think the second point could be an issue for us because mod names (and consequently, the URLs) never change. But as of this writing, "updating KES" essentially requires clicking the button and reinstalling the script, so I think point 1 is being satisfied. So...I don't think any of this should be a problem, it's just that we have to satisfy the twin objectives of fixing the CSS bug while also enabling the different monkey flavors to handle updating correctly.

Pamasich:
I see, we'll have to see how this goes then.

> I belive the @requires will never auto-update on monkey's side (falling back to the locally-stored resources) unless:

Tampermonkey does allow the user to specify the @requires should be updated in regular intervals or even every time the userscript gets executed, but that seems to be a general setting for all userscripts. Not sure if the same functionality also exists on a per-userscript level and in other monkeys.

aclist:
> I see, we'll have to see how this goes then.
> 
> > I belive the `@requires` will never auto-update on monkey's side (falling back to the locally-stored resources) unless:
> 
> Tampermonkey does allow the user to specify the `@requires` should be updated in regular intervals or even every time the userscript gets executed, but that seems to be a general setting for all userscripts. Not sure if the same functionality also exists on a per-userscript level and in other monkeys.

I don't think we've ever tried to atomically update a single mod over-the-air on the stable branch, so it shouldn't be an issue. I remember a lot of issues at the beginning due to people's different usage environments and monkey implementations. E.g., Mac with legacy GreaseMonkey, Windows with ViolentMonkey, and other esoteric combinations. I believe this is why we decided to make updating KES be a monolithic overhaul step, and to the extent possible try to discourage users from using upgrade buttons internal to their chosen extension.

By the way, when using words prefixed with the `@` symbol (e.g. `requires`), it will inadvertently ping a user on GitHub who has that name (I've made this mistake before and did so above), so it's best to wrap them in a code block so that they don't trigger, otherwise it will mention the user. I've edited the comments above to sanitize them.

Pamasich:
> By the way, when using words prefixed with the `@` symbol (e.g. `requires`), it will inadvertently ping a user on GitHub who has that name (I've made this mistake before and did so above), so it's best to wrap them in a code block so that they don't trigger, otherwise it will mention the user. I've edited the comments above to sanitize them.

Oh, thanks for the explanation. Didn't know that, but it makes sense.

Pamasich:
It seems like several urls are actually wrong here, I'll make a fix for that in a bit.

PR #254, opened 2024-03-13
[Mod] Fix Pagination Arrows

Pamasich:
The pagination is currently bugged on some kbin pages. The All Content view and profiles are for sure affected, but this mod works on any others too.

The button to go to the previous page is always disabled and the button to go to the next page always links to the second page. This is regardless of which page the user is actually on.

This mod fixes the behavior and makes them act as intended. 

[The bug was originally reported on the kbin repository.](https://codeberg.org/Kbin/kbin-core/issues/1343)

## Checklist

Add-ons (mods):


Pamasich:
> It seems to be working well; are there any implications for mbin?

The issue doesn't exist on mbin, so this mod just doesn't do anything there. It won't break anything. I'll add a note to the description if there's not already one, similar to my other mods, once I'm back to working on KES.

aclist:
Related: https://kbin.social/m/kbinMeta/t/923110/A-minor-thing-impacting-usability#comments

PR #261, opened 2024-03-14
[alt_all_content_access] Mbin "support"

Pamasich:
A quick fix of the mod for mbin. I've tested this version on kbin.run and it works there.

mbin doesn't currently have a combined threads and microblog view, so this mod is currently useless there. It didn't check for that though before changing the links to point to the All Content address (which doesn't exist on mbin). So I've added the missing check here.

This also includes a general redesign of the mod I did before you encapsulated it in a function, so I've got to ask, did you move the comment at the top inside the method for a reason or only because it was on the class originally? I have it on the function here and I'm not sure if I need to move it inside too.

Also, I've added a note to the description of the mod about it being useless on mbin. I'd like to add a new line there, can I just use `\n` or `<br/>` or is there something else I need to use? I'll check in a bit myself how other mods did it, but figured if I'm writing this anyway, can't hurt to include the question here.

Creating it as a draft for now until I've added the newline.


aclist:
> This also includes a general redesign of the mod I did before you encapsulated it in a function, so I've got to ask, did you move the comment at the top inside the method for a reason or only because it was on the class originally? I have it on the function here and I'm not sure if I need to move it inside too.

I think I just did it mechanically (inadvertently) while consolidating all of the mods in sequence, no special reason here other than my brain wanting to put the function definition at the top. You can put comments wherever you want.

> Also, I've added a note to the description of the mod about it being useless on mbin. I'd like to add a new line there, can I just use \n or <br/> or is there something else I need to use? I'll check in a bit myself how other mods did it, but figured if I'm writing this anyway, can't hurt to include the question here.

I don't think anyone has tried this before, so looks like you are about to find out what happens. In general, using raw HTML inside of mod labels is discouraged, though.

Pamasich:
I tested it just now by editing the cached manifest file and `\n` seems to work.

PR #262, opened 2024-03-14
fix: fixed some broken links (related to #252)

Pamasich:
Forgot to remove the /raw on some of the changed links


aclist:
Yeah, this is going to keep breaking unless I update the KES generation script to stop inserting those suffixes. Will do that now.

aclist:
Just to clarify, I don't think you forgot to update those, they just got re-inserted when the script ran.

Pamasich:
> Just to clarify, I don't think you forgot to update those, they just got re-inserted when the script ran.

Yeah, you're right. Those links were replaced in a merge I did from testing.

aclist:
Should be fixed on origin now if you want to compare the current format with the expected output.

Pamasich:
Oh yeah, looks good now.

aclist:
Thanks.

PR #263, opened 2024-03-14
[fix_codeblocks] mbin

Pamasich:
Mbin doesn't need the mod, Lemmy code blocks work there. The mod isn't causing any issues either.

So I just added a note to the description and didn't make any other changes.


Pamasich:
Ignore the other commits, they're mostly irrelevant now.

Also, TIL I can't mention decorators in commit messages or it'll mention someone. Don't think I can change that, so sorry todo.

PR #266, opened 2024-03-26
[Mod] Unsanitize CSS (another kbin bugfix)

Pamasich:
Kbin currently allows magazines to declare custom CSS that is applied while viewing them, and users can also change the CSS of the site in their settings.

This behavior is currently bugged however. Any CSS defined this way gets HTML escaped, causing certain characters like > and & to be replaced with HTML escape codes. This causes any CSS rule that uses them to not actually work, causing confusion and imposing unnecessary limitations on what can be achieved.

Until kbin fixes this bug themselves, I've implemented this fix as a mod.

I doubt this has a lot of use on its own currently, especially with how underused CSS is on kbin in the first place. And anyone who tests their code wouldn't have CSS that needs this mod running on their magazine anyway. For user CSS, I recommend userstyles over kbin's builtin functionality anyway.
This is very much a "just in case" mod with the reason I'm implementing it now being to act as a dependency (so a use case for the discussed depends_on, once that's implemented) for another bugfix mod which needs to know which style tag contains the magazine CSS.

Add-ons (mods):

### DIFF
--- a/mods/unsanitize_css/unsanitize_css.json
+++ b/mods/unsanitize_css/unsanitize_css.json
@@ -1,0 +1,13 @@
+{
+    "name": "Magazine CSS Fix — Unsanitize",
+    "namespace": "unsanitize-css",
+    "author": "Pamasich",
+    "version": "1.0.0",
+    "label": "Unsanitize Magazine CSS",
+    "desc": "Kbin supports custom CSS for magazines, but that functionality is limited because kbin wrongly replaces some important CSS characters with HTML escape codes, making the code partially non-functional. This addon fixes the issue by undoing those replacements, making the CSS actually work as expected.\n\nIt's unclear if this addon is needed on mbin, but it won't cause any issues if enabled.",
+    "login": false,
+    "recurs": false,
+    "entrypoint": "unsanitize_css",
+    "page": "general"
+  }
+  
--- a/mods/unsanitize_css/unsanitize_css.json
+++ b/mods/unsanitize_css/unsanitize_css.json
@@ -4,7 +4,7 @@
     "author": "Pamasich",
     "version": "1.0.0",
     "label": "Unsanitize Magazine CSS",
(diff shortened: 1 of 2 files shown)


aclist:
Is there a magazine using custom CSS you can link to?

Also, I was thinking of reorganizing the mod pages a bit and putting this and the repair codeblocks under "debug," does that seem reasonable?

Pamasich:
> Is there a magazine using custom CSS you can link to?

Yeah, [mine](https://kbin.social/m/pamasich).

The magazine title in the sidebar is supposed to be red as a test, but it doesn't work because I used `>` as a selector and that gets escaped to `&gt;`. I first ran into the issue when I tried to help someone implement a magazine banner, which was way too much of a struggle to get working.

Pamasich:
> with the reason I'm implementing it now being to act as a dependency (so a use case for the discussed depends_on, once that's implemented) for another bugfix mod which needs to know which style tag contains the magazine CSS.

Actually I just realized that won't work like I was expecting it to... since it's not guaranteed magazine css uses `>` or `&`... So this isn't a guaranteed way to identify the magazine css style element.

Pamasich:
> Also, I was thinking of reorganizing the mod pages a bit and putting this and the repair codeblocks under "debug," does that seem reasonable?

Hmm, I think debug is a bit too technical a term, there's the danger that users won't look in there. I think taking the name literally it makes sense, but usually debug stuff is only meant to be used by the developer, so people who know that would probably not look in there unless they're developing their own mod.

aclist:
The banner is a big red rectangle for me (with no mod applied), is that correct?

Maybe there should be a "Fixes" page. I feel like General is starting to get too cluttered with things that aren't exactly general in nature.

Pamasich:
> The banner is a big red rectangle for me (with no mod applied), is that correct?

Yeah, that's what I ultimately managed to throw together after other attempts failed because of this issue.

Pamasich:
> Maybe there should be a "Fixes" page. I feel like General is starting to get too cluttered with things that aren't exactly general in nature.

Yeah, I've been wondering if there's a better place for them too. A new Fixes category makes sense.

Issue #273, opened 2024-04-08
[BUG] Unintended interaction between Resize Text mod and Collapsible Comments mod


aclist:
If the Resize Text mod is on, toggling the Collapsible Comments mod on/off triggers a timeout and opacity check on the KES modal, because this behavior of the Resize Text mod resides in the global scope and the Collapsible Comments change triggers recursion throughout the page

aclist:
Provisionally resolved by [02440f3](https://github.com/aclist/kbin-kes/commit/02440f3d159360ed9dfbceddc601914b0f7ad2ec), but Collapsible Comments causes the entire DOM for `#comments` to mutate, triggering all mods to iterate and apply. Resize Text now explicitly checks if the KES modal is on the Resize Text mod before applying opacity, but some observations about this side effect:

- Collapsible Comments is slow to set/unset because it's actually iterating over every mod. Any way to add an exception for this (using more granular selectors)?
- Maybe a mod changing the state of the open KES modal should be undefined behavior. Mods shouldn't be manipulating the KES modal itself. Any alternative solution to showing font size increases here? (E.g. a separate preview modal exclusive to the mod?)

aclist:
The opacity issue is resolved by 02440f3, but I still feel there is room for improvement here so that the collapsible comments mod itself does not trigger a mutation that forces recursion through all mods.

PR #276, opened 2024-04-10
[MOD] Change Default Sort

Pamasich:
This mod allows users to set their desired default sort option.

This can be done separately for

- threads & All Content
- microblogs
- comments
- ~~on profiles~~ (support removed for now)
- the list of magazines

When the user visits a page without a sort option specified in the url (like `/m/kbinmeta` instead of `/m/kbinmeta/hot`), they will be automatically redirected to the specified default sort.

This version of the mod works with kbin, but things seem to work differently on mbin at a glance, so there's still some work needed there. I'll do this as a draft for now until that's done.


Pamasich:
Right now I've successfully tested this on the following routes:

- [x] `/`
- [x] `/all`
- [x] `/microblog`
- [x] `/*`
- [x] `/magazines`
- [x] `/sub`
- [x] `/sub/microblog`
- [x] `/*/sub`
- [x] `/m/any`
- [x] `/m/any/microblog`
- [x] `/*/m/any`
- [x] `/u/any/c/any`
- [x] `/u/any/c/any/microblog`
- [x] `/u/any/c/any/*`
- [x] `/c/any`
- [x] `/c/any/microblog`
- [x] `/d/any`
- [x] `/d/any/comments`
- [x] `/tag/any`
- [x] `/u/any`
- [x] `/u/any/threads`
- [x] `/u/any/comments`
- [x] `/u/any/posts`
- [x] `/u/any/boosts`

I think that's all the pages with sorting options on kbin? mbin seems to have some more though.

Pamasich:
Right, I forgot comments

Pamasich:
Mbin looks like it behaves VERY differently. The URLs and everything is completely different.

Pamasich:
Okay, I've assessed what needs to be done for mbin.

- [ ] `/`
- [ ] `/all`
- [ ] `/sub`
- [ ] `/m/any`
- [ ] `/m/any/t/any`
- [ ] `/m/any/t/any/any`
- [ ] `/m/any/t/any/any#comments`
- [ ] `/magazines`
- [ ] `/d/any`
- [ ] `/d/any/comments`

The profile and tag pages don't seem to have any sort options on mbin, so I'll have to add an explanation for that in the description. The current code should abort if no sort options are present, so no adjustment is needed there.

On mbin, sort options are in a dropdown. I'll have to adjust the CSS the mod uses to find them.

The Microblog link in the navbar is relative in mbin, so it always links to the same sort options as is chosen for the Threads portion. In fact, accessing just, for example, `/sub/microblog` will automatically redirect to the full url using `/hot/` and other default options.
For now, I'll leave the microblogs alone. This means they'll stay relative / default to hot if accessed from a page without sort options.
A future improvement for the mod could be an optional functionality to replace the microblog link in the navbar with the desired sort option.

aclist:
Maybe but the mbin disclaimer next to the relevant option labels?

Pamasich:
> Maybe put the mbin disclaimer next to the relevant option labels?

Right, that's better than putting it in the mod's description.

aclist:
I think we could also use the path tokenization and page type enum logic you have here as a generic function other mods can call, because there are a lot of mods that need to check the page type and bail out accordingly (for example, Magazine Instance Names), and it would facilitate making mods (and make the checking logic less brittle) if mods could just call this function and get the page type back without having to reimplement this every time.

Pamasich:
Hmm, I see your point. I'll keep things as they are for now for this mod, but it's a good idea to generalize that logic in the future.

aclist:
It's not something you need to change here, I think I'll just borrow some of your logic and start making a utility function, although I'll probably need help identifying all of the page routes, since I'm not as familiar with the variety of page types as you.

Pamasich:
@aclist I just wanted to use the getInstanceType() function and took a look at how it works.

`const keyw = document.querySelector('meta[name="keywords"]').content.split(',')[0]`

kbin.run, which is the instance I'm using, runs mbin and has these keywords in that element:

`kbin, mbin, link aggregator, microblog, socialmedia, reddit, twitter, open source, fediverse, threadiverse`

As you can see, the code should pick kbin there despite the site running mbin. So I don't think this is as reliable a solution as expected. There seems to be some customizable element to it.

Looking through the first page of mbin results on fedidb, besides kbin.run, there's also:

- rimworld.gallery has "Rimworld Gallery" as its first keyword.
- kbin.chat doesn't even include mbin in the list despite running it
- same for thebrainbin.org
- moist.catsweat.com has "public fediverse instance" and later "mbinorg" but no mbin itself
- gehirneimer.de is the same as kbin.chat
- kbin.life is the same too

This means the current function only works for 3 out of mbin's top 10 instances (and not for the second biggest instance).
Granted, those aren't officially supported by KES currently. But if that's all we cared about, we could also just decide kbin vs mbin hardcoded based on which instance the user is on.

Pamasich:
There is the mbin footer:

image

aclist:
> There is the mbin footer:
> 
> image

It looks like a few instances opt to not include the "About Mbin" footer. But from what I checked so far, the majority include the "Create your own instance" link to https://github.com/MbinOrg/mbin.

`Copy Mbin URL` still seems pretty consistent here, since I can't see why an instance would customize this label beyond the default, but I may be wrong.

aclist:
I'll write a script to poll the top ~100 instances and test whether these elements are present.

Pamasich:
I can tell you that the Create your own instance element isn't on kbin.run.

> `Copy Mbin URL` still seems pretty consistent here, since I can't see why an instance would customize this label beyond the default, but I may be wrong.

I think it's important to keep in mind though, if you end up using that one, that different languages will have it renamed. So the check should probably be if it contains mbin, not if it's literally that string.

Which also reminds me that one of the selectors I use for the sort options probably won't be compatible with other languages too, since I'm using the aria-label there and I assume that changes when the language changes.

Pamasich:
I think the easiest and most consistent would be to just use ActivityPub.

When requesting the instance itself (`kbin.social` or `fedia.io` for example) with an `Accept` header of `application/activity+json`, the result will be a json object with information about the instance.

This json includes a field `name` which has a value of `Mbin` for mbin and `kbin` for kbin.

Pamasich:
If performance is a concern, this could be done once by KES and then stored for when a mod needs the information.

aclist:
Well, to my surprise, there weren't even 100 instances. Anyway, here are the results:

These three domains are unavailable:

- teacup.social
- kbin.dk
- the.coolest.zone

This domain does not have the `.about-mbin` element:
kbin.life

All of the other domains have the `.about-mbin` element.

But what all of the domains do have, whether or not they have the `.about-mbin` element, is a link to the MbinOrg GitHub organization. That's located in either the create your own instance area, the about area, or both, but every domain at least had this URL once on the page. So whether or not some element is on the page, the string `MbinOrg` should be present.

> I can tell you that the Create your own instance element isn't on kbin.run.

I didn't find this to be the case, it is right there on the page for me and in the source. Is KES hiding this element?

aclist:
> I think the easiest and most consistent would be to just use ActivityPub.
> 
> When requesting the instance itself (`kbin.social` or `fedia.io` for example) with an `Accept` header of `application/activity+json`, the result will be a json object with information about the instance.
> 
> This json includes a field `name` which has a value of `Mbin` for mbin and `kbin` for kbin.

Makes sense as well, the mod can just look up this variable. Although it seems a bit unfortunate that we can't easily get the response headers for the current page with JS, necessitating making a second query.

Pamasich:
> I didn't find this to be the case, it is right there on the page for me and in the source. Is KES hiding this element?

Yeah, looks like something is applying `display: none;` to it. And it's not my userstyle.

image

aclist:
> Yeah, looks like something is applying `display: none;` to it. And it's not my userstyle.

Presumably `Explore the fediverse intro text`?

Pamasich:
Ah, it's the `Hide sidebar elements` mod. I forgot that's a thing.

aclist:
> Ah, it's the `Hide sidebar elements` mod. I forgot that's a thing.

Sorry, I meant to say the `Create your own instance promo` text. Brain is not working today.

aclist:
Performance-wise, I think this is going to be faster than making an XML request to the page again, particularly given that some instances are slow to load:

```
function getInstance(){
    const links = document.getElementsByTagName("a");

    for (var i = 0, l = links.length; i < l; i++) {
      const link = links[i];
      if (link.href.indexOf("MbinOrg") !== -1) {
        console.log("mbin");
      }
    }
    console.log("kbin")
}
```

If this proves to be an issue in the future, we can rethink this. But I'm actually at the point where I want to reduce performance bottlenecks on KES, because I feel like certain setup and mod application steps are consuming unnecessary cycles. Particularly, for example, #273

Pamasich:
@aclist Does KES currently intend to support other languages than english? I just realized the mod in its current state is all-around incompatible with other languages since I'm using interface strings, and I'm not sure if I should try fixing that now or in a future update to the mod.

aclist:
It's a valid point that I don't think we've really considered before. Can you point me to a specific part of the code in the mod? Are you talking about aria labels?

Supporting languages at more than the kbin code level would also have to include localizing the KES interface, so it's not a small undertaking.

We definitely need to have an admonition that mods should parse underlying elements and path tokens, not interface strings, where possible.

Are languages configured in the kbin settings at the user's preferences, or are they defined by the instance owner and apply to the entire instance?

Pamasich:
> Can you point me to a specific part of the code in the mod? Are you talking about aria labels?

Basically the entirety of the `makeOptionsExplicit` function uses this repeatedly: `option.textContent.toLowerCase().trim()`.

> Are languages configured in the kbin settings at the user's preferences, or are they defined by the instance owner and apply to the entire instance?

The former. The footer (or corresponding element in mbin) has a language selector, the value is stored as a cookie. Regardless of kbin or mbin, the cookie is currently called `kbin_lang` and contains values like `en` or `de`.

aclist:
If I understood the contents of that function correctly, kbin defaults to a certain sort order that doesn't actually reflect the currently active sort element, so, e.g., `hot` is selected but points to `/sub`, whereas the other ones will point to `/sub/newest`, correct? But can't there only be one non-explicit default sort order on a page at a given time, with all of the other sort hrefs actually mapping to valid paths? So wouldn't the odd one out of the five (the one with a simple pathname with no suffix) be the one with the non-explicit sort, regardless of what language its label is in? By process of elimination, it can only be the remaining string, which is always internally in English. Apologies if I'm missing something obvious here.

> The former. The footer (or corresponding element in mbin) has a language selector, the value is stored as a cookie. Regardless of kbin or mbin, the cookie is currently called kbin_lang and contains values like en or de.

Thanks. Given that there are a finite number of sort elements and set languages, you could also poll the cookie and construct some kind of stringwise mapping, just hardcoding the sort label strings (5 * 17 languages), but it probably wouldn't be necessary if using the approach above.

Pamasich:
> But can't there only be one non-explicit default sort order on a page at a given time, with all of the other sort hrefs actually mapping to valid paths? So wouldn't the odd one out of the five (the one with a simple pathname with no suffix) be the one with the non-explicit sort, regardless of what language its label is in? By process of elimination, it can only be the remaining string, which is always internally in English. Apologies if I'm missing something obvious here.

That's true, but then I have to change the link so it points to the explicitly sorted url. For which I'm currently using the text of the option.

Pamasich:
I guess I do have an idea for how to do it, based on what you suggested.

aclist:
I think you could just use a simple array of strings or object mapping here to define the possible options upfront, so if, after processing the sort options with valid suffixes, the remaining link is none of the above, you set it to the remaining option. If the remaining option is "newest" but the label is "banana", it shouldn't matter.

Pamasich:
@aclist I tried using the bash script you wrote to add the mod to the funcs.js file and it gave me a bunch of errors about the `sponge` command being missing. It seems it correctly added the mod to the file though. Can I ignore the errors then, or does sponge do something important?

Pamasich:
I'm not sure if this mod is causing them, but trying to test the mod out with KES, I'm noticing random delays in getting redirected, caused by waiting for the bug report and collapsible comments mods which keep getting errors a bunch of times.

I didn't notice any issues like this in the past, so it's plausible it's because of this mod. But I can't think of why it would be.

image

There's a wall of these, the above are just two examples.

aclist:
> @aclist I tried using the bash script you wrote to add the mod to the funcs.js file and it gave me a bunch of errors about the `sponge` command being missing. It seems it correctly added the mod to the file though. Can I ignore the errors then, or does sponge do something important?

sponge is part of the moreutils package and is used to edit a file in place by writing the contents of the stream into the same file. Usually, you cannot write the contents of a pipeline into the selfsame file, because the redirection operator at the end of the pipeline will be read first, truncating the file before you can process the stream at the left-hand side. Effectively, sponge just creates a temporary file to soak up the data. It's just a convenience tool, but it slipped my mind that it wouldn't be available out of the box. I'll update the script to just write to temporary file and then move it back to the desired location. Without it, I don't think you're going to get the desired output.

aclist:
> ~I'm not sure if this mod is causing them, but trying to test the mod out with KES, I'm noticing random delays in getting redirected, caused by waiting for the bug report and collapsible comments mods which keep getting errors a bunch of times.~
> 
> ~I didn't notice any issues like this in the past, so it's plausible it's because of this mod. But I can't think of why it would be.~
> 
> image
> 
> ~There's a wall of these, the above are just two examples.~
> 
> Never noticed this before, but apparently it's got nothing to do with this mod.

It's probably because the funcs.js file is malformed as a result of the script not behaving correctly.

aclist:
I updated the `concat_funcs` script. It doesn't use sponge anymore, and takes one optional argument: `-g`. If this flag is enabled, it will also add the resulting files to the git index using `git add`. Otherwise, it will just update the funcs file and manifests without explicitly adding the changed files to the git index.

aclist:
Incidentally, there is a piece in the documentation that describes how to use this in conjunction with the gist generator: https://aclist.github.io/kes/kes.html#_prototyping_local_code

This is working for me and saves time, but I'll build it up to be a bit more user-friendly.

Pamasich:
I'm not sure if I can work on this this week.

aclist:
> I'm not sure if I can work on this this week.

Not a problem.

Pamasich:
Well, seems like one week became three, probably four.

Work has gotten more busy and I've got some more momentarily important programming projects currently which are blocking me from working on KES.
I'll hopefully be able to return to this within the next two weeks.

Pamasich:
@aclist 
I'm trying to build the mod right now so I can test out the options. I've noticed that the current manifest generation script produces an empty file for me. Putting quotes around the *.json fixes that and produces the correct file. It's probably another issue from me using Windows.
Would you mind changing the script accordingly? Or I guess, would you mind if I committed my change to it.

Pamasich:
> Sure, is this what you wanted?
> 
> ```diff
> < for i in $(find $PWD/mods -name *.json); do 
> ---
> > for i in $(find $PWD/mods -name "*.json"); do 
> ```

Yeah, that's exactly it.

aclist:
OK, done.

Pamasich:
The mod works for me now, tested on kbin.social and kbin.run.

Support for profile pages is out for now, because kbin.social didn't work when I implemented most of it, and kbin.social is the only kbin instance with sort options on profiles anyway. So it didn't feel worth it enough to go back and edit it back in. I'll add it back in the future though.

I'll make some more cosmetic changes to the code, like adding some more comments and reordering some stuff. But other than that I do consider it finished now.

aclist:
Alright, thanks. I'll start taking a look through it tomorrow.

Pamasich:
Okay, I've completed the changes I still wanted to make and did another quick test run to check that it didn't break anything.

Pamasich:
Discovered a bug with this: the mod currently breaks when there's an url parameter (which is the case when trying to change the sort order on any except the first page)
I'll push a fix for this bug tomorrow.

Pamasich:
Tested the fix, it works now.

Pamasich:
There might actually still be another bug, for some reason I'm getting "/hotundefined" links now on kbin.social with this mod. I'll have to investigate tomorrow why that is.

Pamasich:
To clarify: it was an oversight with my previous fix, where I had falsely assumed an undefined value would be stringified as an empty string rather than "undefined".

aclist:
I think I found some bugs here, but my testing methodology might be flawed.

1. If I visit https://kbin.run/m/announcements and explicitly enable this mod and set the thread index sort method to `default`, then reload the page, it tries to navigate to https://kbin.run/m/announcements/hot, presumably because `hot` is the default sort method, but this URL does not actually exist.

2. If I go into a thread and explicitly set a sort method, such as `newest`, it works accordingly, and this is appended to the URL path. If I subsequently change the sort method, or change the sort method and then attempt to reload the page in place, the URL does not change to the new sort method, staying on `newest`. I have to manually go to the bare URL (with no sort method appended) for the regex replace to happen. So as a side effect of this, it seems that changing sort methods on the fly without manually scrubbing the URL is not working.

3. As a clerical matter, I found "Select default sort method for magazines" to be a bit confusing, as I didn't initially realize it referred to the magazines table. I thought it referred to the magazine index (the thread index) for a particular magazine. Somewhat similarly, I feel like "threads" and "comments" are liable to be confused here, so maybe there is slightly better wording? I get that the explanation is stating "what is being sorted", but it seems like "where is it being sorted" is the typical frame of reference someone uses. Of course, there might be standalone threads and/or comments in some other places not specified by the above, so I understand the need to state "what is being sorted", but it took me a bit of fumbling to find the right options, so maybe a parenthetical explanation is needed here:

- Select default sort method for threads (top level posts on a magazine's thread index)
- Select default method for comments (posts by users inside of a thread)
- etc.

Pamasich:
> If I visit https://kbin.run/m/announcements and explicitly enable this mod and set the thread index sort method to default, then reload the page, it tries to navigate to https://kbin.run/m/announcements/hot, presumably because hot is the default sort method, but this URL does not actually exist.

Huh, the same exact scenario works on the frontpage, which also ends in /threads, so I think I just assumed it works the same in magazines. I'm sure I specifically tested magazines though, was that maybe only on kbin...

Well, good catch, I'll have to look into fixing that.

> If I go into a thread and explicitly set a sort method, such as newest, it works accordingly, and this is appended to the URL path. If I subsequently change the sort method, or change the sort method and then attempt to reload the page in place, the URL does not change to the new sort method, staying on newest. I have to manually go to the bare URL (with no sort method appended) for the regex replace to happen. So as a side effect of this, it seems that changing sort methods on the fly without manually scrubbing the URL is not working.

I'm not sure I'm understanding this one correctly.
With "If I subsequently change the sort method", do you mean in the KES settings?

It's by design that that only applies when visiting the non-explicitly sorted page. Not only do I think it would be annoying to have it redirect you just from changing the settings, I also don't know if it's even possible without locking you out of any sort options but the default one.
Those settings are just meant to tell where to bring the user when no sort option is explicitly specified in the URL, which isn't the case on a /newest URL.

> As a clerical matter, I found "Select default sort method for magazines" to be a bit confusing, as I didn't initially realize it referred to the magazines table. I thought it referred to the magazine index (the thread index) for a particular magazine. Somewhat similarly, I feel like "threads" and "comments" are liable to be confused here, so maybe there is slightly better wording? I get that the explanation is stating "what is being sorted", but it seems like "where is it being sorted" is the typical frame of reference someone uses. Of course, there might be standalone threads and/or comments in some other places not specified by the above, so I understand the need to state "what is being sorted", but it took me a bit of fumbling to find the right options, so maybe a parenthetical explanation is needed here

Hmm, I see your point with the magazines one, I should be more clear there. But the issue with threads and comments is that they're not just used in one place. Off the top of my head, domain pages have both threads and comments, and the threads option is also used on the frontpage. So your suggestions wouldn't really fully work. I do think the wording could be made a bit more specific to make sure people understand threads are for thread indexes/overviews, not individual threads, and the comments option refers to comments on threads, not microblogs. But I do think, besides those two points, that those two options are plenty clear already. The threads pages are clearly labeled in the UI, and everyone knows what a comment is.

I think I'll change them to something like (just a rough in the moment idea, going to think more about it until I return to work on this):

- Choose by what to sort the list of magazines when no option is specified
- Choose by what to sort lists of threads when no option is specified
- Choose by what to sort comments of threads when no option is specified
- Choose by what to sort microblogs when no option is specified

This doesn't feel right yet though, I'll need to think more about this. I don't like how long they are like this, so maybe I'll also move part of the explanation to the end of the description instead.

What are your thoughts on those?

aclist:
> With "If I subsequently change the sort method", do you mean in the KES settings?

Hmm, I think this one is on me. Using KES settings, not the sort on the page. I must have been tired when I experimented with it and for some reason expected changing the default sort would re-sort "in place," so to speak. My expectation was that it would take you to the non-explicitly sorted page, after which the new default would apply. But for this to work, the mod would have to also evaluate whether you are on a page that corresponds to the filter being changed, and there is probably a bad side effect to that.

> It's by design that that only applies when visiting the non-explicitly sorted page. Not only do I think it would be annoying to have it redirect you just from changing the settings, I also don't know if it's even possible without locking you out of any sort options but the default one.

Right, I see your point.

It's been awhile since the initial PR, so I tried to jump into this mod as blindly as possible just to see what the initial user experience would be, so I admit I didn't read the fine print. I looked at the mod dsecription again and it seems sufficiently self-explanatory.

> I do think the wording could be made a bit more specific to make sure people understand threads are for thread indexes/overviews, not individual threads, and the comments option refers to comments on threads, not microblogs. But I do think, besides those two points, that those two options are plenty clear already. The threads pages are clearly labeled in the UI, and everyone knows what a comment is.

Yeah, my quibble isn't with the terms "threads" and "comments" per se, more so that without additional context, "threads" alone could be read either as "stuff inside of threads gets sorted" or "threads themselves get sorted," but then you come across the "comments" option, which are the "stuff inside of threads," so there can't logically be two of those, so you go back and look at the threads option again and reason it must be threads themselves, but ideally you shouldn't have to go through this thought process, IMO. It's also because kbin hasn't done a good job of disambiguating between posts, threads, articles, etc., so I'm not sure if most users consider threads to be "posts," or whether they refer to them as "threads."

> I think I'll change them to something like (just a rough in the moment idea, going to think more about it until I return to work on this):

    Choose by what to sort the list of magazines when no option is specified
    Choose by what to sort lists of threads when no option is specified
    Choose by what to sort comments of threads when no option is specified
    Choose by what to sort microblogs when no option is specified

> This doesn't feel right yet though, I'll need to think more about this. I don't like how long they are like this, so maybe I'll also move part of the explanation to the end of the description instead.
> What are your thoughts on those?

Seems like the same phrase is being reused on every option, so I would make that part of the description followed by a line break, then add the options below.

```
Choose the default sort order for the following contexts:

The global magazines list
The thread index in magazines
Comments inside of threads
Microblog posts
```

That's merely off the top of my head.

It might be kind of cool to have a link to the magazines list in option 1, but I don't think we support a link right in the option itself. It is possible for each option to have a label (if necessary) above it, though.

Pamasich:
> If I visit https://kbin.run/m/announcements and explicitly enable this mod and set the thread index sort method to default, then reload the page, it tries to navigate to https://kbin.run/m/announcements/hot, presumably because hot is the default sort method, but this URL does not actually exist.

In the newest commit I've changed the approach used to build the URL, so this should work correctly now. Haven't thoroughly tested the new code yet though, so I don't know if it breaks anything else.

Pamasich:
I reworked the description and settings of the mod to hopefully be a bit clearer.

The 'default' option now behaves differently. Instead of going with the instance's default sort option, it now disables the mod for that page type. This should avoid unnecessary redirects.

aclist:
It seems to be working well now.

aclist:
One question, the description for the mod refers exclusively to Mbin now. Does this imply that the mod is not intended to work on Kbin? Or were you just using the term Mbin as a generic term since Kbin is kind of obsolete now?

Pamasich:
> One question, the description for the mod refers exclusively to Mbin now. Does this imply that the mod is not intended to work on Kbin? Or were you just using the term Mbin as a generic term since Kbin is kind of obsolete now?

The latter. While I'd love to see ernest return, kbin.social looks pretty dead and other instances have pretty much gone as well. Even kbin.cafe, which has held out for pretty long, seems to be offline now. There's not going to be any kbin users using this script, so no need to mention it imo.

aclist:
Time to consider a name change for this tool? I also need to scrub the link to kbin.social in the footer of the KES modal.

PR #295, opened 2024-06-14
[MOD] Alternate Spoilers

Pamasich:
## Explanation

Spoilers on kbin/mbin work a bit differently than I'm used to from other platforms. Instead of kbin's `<details>` approach, they're usually displayed as obscured blocks that show their contents when hovered or clicked on.  
This mod aims to switch out kbin's spoilers with ones closer to the ones seen on other platforms.

With this mod, spoilers now display as blocks with only their title visible, which reveal their content when hovering or when the title row is clicked (they can be hidden again by clicking again). Both of these means of interaction can be toggled by the user in KES.

## Images

image
*Native spoilers (expanded)*

image
*New spoilers when not hovered or clicked*

image
*New spoilers while hovering over them or after they're clicked on*

## Todo (before 1.0)

- [x] Test the mod on mbin
- [x] Test the mod with KES
- [x] Change the cursor when hovering over the title while showOnClick is enabled
- [x] Documentation (comments)
   - [x] Add a user-facing explanation of how clicking works (only clicking on the title row does show/hide the spoiler contents)

Edit: I probably should eventually improve the design, like making it more different from code blocks. But for now this should be good enough for a 1.0.


aclist:
Is it possible to spoiler individual words inline, or only chunks of text? (Sounds like the latter for now?) For example, see this page as a reference: https://steamcommunity.com/app/2458980/discussions/0/6811212011865688045/

Pamasich:
> Is it possible to spoiler individual words inline, or only chunks of text? (Sounds like the latter for now?) For example, see this page as a reference: https://steamcommunity.com/app/2458980/discussions/0/6811212011865688045/

No, sadly kbin doesn't have support for inline spoilers. Which is a shame. But ernest essentially just added support for Lemmy's spoiler syntax, and they happen to not support inline spoilers either.

Inline spoilers could be added using the old trick used by some subreddits (before Reddit added their native spoilers) of doing a custom link syntax like `[spoiler title](/s "spoiler content")`. But I doubt that would be super useful if only KES users could read those spoilers.

aclist:
I see, I see.

aclist:
Getting around to testing this now, but spoiler creation seems to be undocumented and not provided OOTB in the text (k|m)bin editor. Do you have a link to a post with spoilers, or have some additional instructions on the invocation to use to create them?

Pamasich:
> Getting around to testing this now, but spoiler creation seems to be undocumented and not provided OOTB in the (k|m)bin text editor. Do you have a link to a post with spoilers, or have some additional instructions on the invocation to use to create them?

```
::: spoiler BUTTON TEXT
HIDDEN TEXT
:::
```

is the spoiler syntax used on lemmy (and adopted by kbin/mbin)

aclist:
I am wrapping up another project and will be reviewing/approving these this week.

aclist:
Tested it and the implementation looks really nice.

I think there is an unintended interaction between the "hover" and "click" actions when both are enabled. The click action takes precedence and hovers stops producing any effect on a clicked spoiler.

Example, a spoiler with content that is not currently visible:

- With both actions enabled, hover the spoiler and then move the cursor outside it
    - As expected, the content shows and then hides

- With both actions enabled, hover the spoiler and then click the title once
    - The content shows on hover, and clicking triggers no action, because the spoiler is already shown and the initial click action is supposed to "show" the spoiler
    - The expectation here (I assume) is that clicking on an already hovered spoiler title would hide it

- Having now clicked the title, move the cursor outside of the spoiler area
   - The spoiler does not hide, because the click behavior is overriding the hover action
   
- Now return the cursor to the spoiler area and click the title
    - No visible effect. This has merely "turned off" the click action
    - Move the cursor outside of the spoiler to see that the spoiler content hides, since only the hover action is enabled now

Pamasich:
It was actually my intention that that's how it works. But I can see how it's confusing. I'll think about how to make it more clear how things work.

Like, clicking and hovering are currently presented as two different methods of showing the spoiler, but they actually have different purposes.
Clicking isn't actually supposed to show or hide the spoiler, but rather to toggle whether it should be hidden in the first place.

> With both actions enabled, hover the spoiler and then click the title once
> - The content shows on hover, and clicking triggers no action, because the spoiler is already shown and the initial click action is supposed to "show" the spoiler
> - The expectation here (I assume) is that clicking on an already hovered spoiler title would hide it

When you hover the spoiler, it's still set to the default behavior of hiding again as soon as the cursor leaves it. Clicking it is meant to toggle that behavior, not whether the spoiler is shown. So considering that behavior is still active, clicking it disables it, causing the spoiler to be shown permanently.

I think this could be better explained in the mod itself, and I should add to my post-1.0 plans to add some visual feedback to clicking so people know it actually does something. Maybe add an icon that shows the spoiler is set to show at all times. But I probably won't add that one to this version yet.

---

Tl;dr: This is intended behavior, but I'll think about how to make it clearer.

aclist:
It seems like these should have their own (separate) heading labels and not be grouped in the same category, then, and the label for the click action be clarified as being a toggle that applies to the hover action.

I think a radio button might be the better choice in order to indicate that these are independent actions, rather than a multiple choice.

Just paraphrasing below for the sake of example:

```
Description: shows spoilers in block format

Hover to show/hide spoiler:
⦿ ON  ◯ OFF
Clicking on spoiler toggles visibility (always show/auto hide):
⦿ ON  ◯ OFF
```

Also spitballing here, the second option could maybe be a permanent setting that doesn't require clicking? Just set ahead of time with this radio button whether spoilers are perma visible or auto-hide when hovered off. Just an idea.

mods/blocky_spoilers/blocky_spoilers.json
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "Reddit style spoilers",
+    "namespace": "blocky-spoilers",
+    "author": "Pamasich",
+    "version": "1.0.0",
+    "label": "Alternate spoilers",
+    "desc": "Spoilers on kbin/mbin are implemented as an element that can be collapsed and expanded. This is different from the usual approach where the spoiler text is shown in full but obscured by a black bar until hovered over or clicked.\n\nThis addon changes spoilers to work like on other platforms, which you might be more familiar with.\n\nBy default you'll be able to toggle whether a spoiler is obscured by clicking on its header (the always visible text). Optionally, hovering is also supported to temporarily show a spoiler. But you'll still have to click to make it stick.",
```
aclist, line 7: "But you'll still have to click to make it stick." The subject of the prior sentence is "hovering", but the subject of this sentence is "the spoiler". Also, "by default" seems unnecessary here, since there is no alternate mode for the click behavior itself, and hovering doesn't itself modify the clicking.

Suggested wording based on my understanding (also rearranged some clauses):

Click a spoiler's header (the always visible text) to toggle whether the spoiler's content should be permanently visible (sticky). This behavior is off by default. Click once to turn on, then click again to turn off. Optionally, hovering is also supported to temporarily show/hide spoiler content, provided the spoiler is not currently set to sticky.

aclist:
Thoughts on the following cosmetic changes?

- Change spoiler title color on hover (just picked a generic color, should use kbin internal theme vars)

a

- Show lock icon (`<i class="fa-solid fa-lock"></i>`) when spoiler is set to perma visible (sticky)

b

Pamasich:
> Thoughts on the following cosmetic changes?
> 
> * Change spoiler title color on hover (just picked a generic color, should use kbin internal theme vars)
> 
> a
> 
> * Show lock icon (`<i class="fa-solid fa-lock"></i>`) when spoiler is set to perma visible (sticky). Hopefully this ameliorates the ambiguity of whether the option is on, and avoids the need for fancy status animations
> 
> b

Thanks for the suggestions.

The design here isn't meant to be final, there's still quite some work to do on it. I've just pushed them to after a first release, as in my opinion the functionality is more important than having it look perfect.

I was personally thinking of displaying a transparent eye icon on hover that becomes solid when clicked. Your suggestion of a lock does make a lot more sense especially when hover is enabled, but I think it's less clear when hover is disabled. Do you think using both an eye or a lock depending on whether hover is enabled or disabled would be confusing to the user?

As for the color, I'm not sure if that would be the best idea. I feel like that would draw unnecessary attention to the spoiler title. Making it clearer that it can be clicked is one of those areas where I was planning to improve the design post-release. But I'll want to think of other options before implementing color.

mods/blocky_spoilers/blocky_spoilers.json
Pamasich: > "But you'll still have to click to make it stick." The subject of the prior sentence is "hovering", but the subject of this sentence is "the spoiler".

I meant it more as the subject being the change to whether the spoiler is visible. But yeah, that's not clear.

> Also, "by default" seems unnecessary here, since there is no alternate mode for the click behavior itself

Yeah, agree.

> Click a spoiler's header (the always visible text) to toggle whether the spoiler's content should be permanently visible (sticky). This behavior is off by default. Click once to turn on, then click again to turn off.

I think this is a bit too ELI5. The "toggle" and "off by default" already implies that clicking once turns it on and then clicking it again turns it off again. I also don't think specifying the default state is necessary, as it's just logical in my opinion that a spoiler would be hidden by default. When I read this, it sounds to me more like the ability to click is off by default.

I'll see what I can do to improve the description. I should also mention the lock icon as an indicator of whether the spoiler is sticky or not once that's implemented, which would further reduce the need to define the default state.

Pamasich:
Before I'll make these changes though, I need to do some work on my third mod that currently has a pull request here, before I forget what I wanted to do there.

mods/blocky_spoilers/blocky_spoilers.json
aclist: Yeah, I agree with that. I was just trying to get the salient points out and list them all one by one, but perhaps I went overboard.

I also agree that the description could greatly be streamlined when cosmetic/visual indicators are added.

aclist:
> I was personally thinking of displaying a transparent eye icon on hover that becomes solid when clicked. Your suggestion of a lock does make a lot more sense especially when hover is enabled, but I think it's less clear when hover is disabled. Do you think using both an eye or a lock depending on whether hover is enabled or disabled would be confusing to the user?

I rewrote a reply to this like five times but couldn't come to a definitive conclusion about what I was thinking. It seems like the main problem to be solved here is that you have multiple modes the state of which isn't always obvious:

- Spoiler is not expanded and hover is disabled: hovering does nothing
- Spoiler is not expanded and hover is enabled: hovering expands the spoiler

- Spoiler is expanded and hover is disabled: regardless of hover setting, spoiler stays expanded
- Spoiler is expanded and hover is enabled: regardless of hover setting, spoiler stays expanded

Overloading these states with too many different icons might be confusing if they don't really relate to each other or don't have a clearly-defined, widely accepted meaning.

It seems like maybe keeping the original close/expand arrow in the new style might solve this issue, for the most part? I feel like some clarity was lost by removing it. It has a universal meaning for closed/open content. If the content is already visible AND the arrow is set to expand (downward pointing arrow), it should be clear at this point why hovering off does nothing--the spoiler is open.

Revisiting the patterns above:

- Spoiler is not expanded and hover is disabled: user can hover over the title and see that hover is off
- Spoiler is not expanded and hover is enabled: user hovers the title and the spoiler physically expands, showing the downward arrow. After hovering off, it rolls up again and shows the right-pointing arrow

- Spoiler is expanded and hover is disabled: the spoiler is forced open already, so hovering on or off of it shouldn't do anything
- Spoiler is expanded and hover is enabled: same as above

The only minor ambiguity this leaves is not knowing at a glance whether hover is turned on in the options. On the one hand, isn't that something the user would have already checked at setup time in the KES settings modal? And it can confirmed by hovering.

On the other hand, if this ambiguity needs to be resolved, then I see two options:

- A set of two icons, either of which is permanently displayed, and indicates whether hover is on or off in the settings
- A single icon which appears on hover

As for the latter, it seems weird that you would have to hover to get an icon indicator of whether hover works. And if the expand arrow is restored, this would probably be superfluous.

As for the icon itself, not sure if eye is the right play here, because if the spoiler is already manually expanded by click but hover is off, then a disabled eye icon seems out of place if the content itself is visible. Maybe some kind of hand pointer? Or maybe just a stylized "H"?

> As for the color, I'm not sure if that would be the best idea. I feel like that would draw unnecessary attention to the spoiler title. Making it clearer that it can be clicked is one of those areas where I was planning to improve the design post-release. But I'll want to think of other options before implementing color.

I was just talking about a rollover effect so that the title looks more like an interactive button/link, because currently it's a bit hard to notice that the cursor changed to a clickable pointer and the title is liable to look like it's part of the content.

Pamasich:
I'll swap the pull request back to draft until the changes are made.

aclist:
I'll be active again starting on 2024-09-05 and wil merge the completed PRs.

Pamasich:
Going to close this for now until I make more changes to it.

PR #303, opened 2025-01-16
fix: Added the clarify_recipient text back to the page

Pamasich:
The mod didn't show the recipient like it claims to do. The reason being that it's element selector didn't work.

I'm not sure where exactly the recipient was actually meant to be added originally, but for now I've changed the code so it gets added just above the message box.


aclist (changes requested):
Working well. I have a slight improvement to make:

```diff
14c14,18
<         form.innerHTML = "Sending message to " + recipientName + form.innerHTML
---
>         const holder = document.createElement('div')
>         holder.id = "mes-compose-holder"
>         holder.innerHTML = "Sending message to " + recipientName
>         holder.style.cssText = 'padding-right:10px'
>         form.prepend(holder)
```

I noticed that the string was abutting the edge of the form without any padding. Rather than inserting the raw string directly into the form's HTML, sticking it in a holder element should make it a bit easier to work with and maintain in the future.

Pamasich:
I'm not quite sure what you mean with this part

> I noticed that the string was abutting the edge of the form without any padding.

It's only ever as wide as the textarea at most, so how is this a problem?

Though I see nothing wrong with the extra padding, it doesn't break anything, so I'll add it.

aclist:
Maybe I am just not accustomed to the mbin "messenger" style of the inbox display; I wasn't entirely sure what I was looking at. With kbin, there was a compose field, and the recipient name would go above that.

With mbin, the compose field is inside of the "conversation", so the recipient name can't go above it. So it made sense to put it to the side, but at least on my setup, the username was completely flush with the compose field with 0 pixels of padding. (I used a very long username as a test).

aclist:
Aha, so we were looking at different pages.

> My latest commit does implement your suggestion and it looks like what you're describing there, but I honestly find it a bit ugly like that. Do we actually need the mod to work on that page though? The name seems to already be given at the top, just without the instance name (which calls more for an expansion of the user_instance_names mod to this page). So if you don't mind, I'll just disable the mod on this page.

I agree, it doesn't look great, and it seems unnecessary on this page.

aclist:
Using https://github.com/aclist/kbin-kes/pull/371, it should now be possible to check if the returned page type of `getPageType()` is `Mbin.Messages.Thread` and abort. The direct message page is `Mbin.User.Direct_Message`.

PR #307, opened 2025-01-16
Fixed issues with Federation Awareness's cleanup

Pamasich:
The mod didn't clean up its CSS properly, leading to two issues:

1. When switching between bubble and border mode, the previous setting wasn't disabled. You simply now get both of them until you reload the page.
2. Only the bubble got cleaned up properly when the mod was turned off. Turning off the mod didn't do anything about the border. Instead, it was removed the next time the mod was turned ON. And so it would alternate between showing the border or not showing it every time it was turned on.

I fixed thsoe two issues by overhauling the cleanup. It now actually makes use of `removeStyle`.

### DIFF
--- a/mods/kbin_federation_awareness/kbin_federation_awareness.json
+++ b/mods/kbin_federation_awareness/kbin_federation_awareness.json
@@ -1,7 +1,7 @@
 {
   "name": "kbin-federation-awareness",
   "author": "@Ori@sacredori.net",
-  "version": "0.2.5",
+  "version": "0.2.6",
   "label": "Federation Awareness",
   "login": false,
   "recurs": true,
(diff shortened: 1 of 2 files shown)


aclist:
Thanks. Looking at the code alone, it seems much cleaner. Haven't tested it fully yet. I think we should have more of these abstracted functions like `removeStyle()`, since there are certain actions that most mods do over and over again.

aclist:
Tested, looks good. This mod uses a lot of "kbin" nomenclature that should probably be scrubbed at some point. It also seems intuitively wrong for "kbin" to be part of the mod name; this deviates from every other mod. On that note, same (probably?) goes for "improved collapsible comments", since there is no additional context here anymore that it's a fork of a standalone script.

PR #308, opened 2025-01-16
[FIX] Notification panel failed to load

Pamasich:
This mod had two issues.

1. The button for the notifications panel wasn't added. This was because of the `.mbin-container` change.
2. The contents didn't load, because the tokens for the mark as read and purge buttons weren't found and so an exception was thrown.

The two commits here fix those two issues.

### DIFF
--- a/mods/notifications_panel/notifications_panel.json
+++ b/mods/notifications_panel/notifications_panel.json
@@ -4,7 +4,7 @@
     "blobcat",
     "shazbot"
   ],
-  "version": "0.3.1",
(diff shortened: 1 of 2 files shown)


aclist:
Just listing things as I see them here, since I still need to mock messaging between users and the read/purge actions to fully test this mod:

- Ideally, the cursor should not be a hand icon if there is nothing clickable in the notifications panel (i.e., "No new notifications!". There is no clickable interaction here. Maybe as an alternative, make clicking the empty panel destroy it.

PR #310, opened 2025-01-16
[FIX] resize_text mod didn't work fully

Pamasich:
Fixed a few issues with some of the options in this mod:

- The description claimed a reload would be required to turn the mod off, but that wasn't the case for me at all.
- The footer option didn't work because Mbin moved its "footer" to the sidebar where it's affected by the regular sidebar option. So I removed the option.
- The sort by option only applied to the icons in the sort bar, not the actual text.
- The notifications and the messages title were suffering from the `.mbin-container` issue.

### DIFF
--- a/mods/resize_text/resize_text.json
+++ b/mods/resize_text/resize_text.json
@@ -1,7 +1,7 @@
 {
   "name": "Customize font size",
   "author": "minnieo",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "label": "Change font size",
   "login": false,
   "recurs": true,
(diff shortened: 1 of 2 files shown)


aclist:
I noticed a potentially bad bug in the mod itself while testing this:

- The user can set an empty value for any of the font size fields. For example, delete the content of the `Notifications` input field. This saves the font size in localStorage to an empty value. 
- Upon opening the MES modal again, any fields under (and inclusive of) that field will fail to load

This could be attacked a few different ways:

- On the mod side, normalize empty font values to a default size. Not sure if this is necessary or possible right now, since I think the bug happens upfront in MES.
- On the MES side: there are cases where it's permissible in MES to set an empty value in an input field. For example, if the field is used to add a text string to some label, the user can opt to have no string at all. However, maybe there's additional handling here where we could explicitly set `input type=number` fields to have special handling: it should be possible to increment their value, but not delete it altogether.

aclist:
> I noticed a potentially bad bug in the mod itself while testing this:

Resolved by da82834decc8ece722dfdce5645f2ff5cca52cc8

PR #311, opened 2025-01-16
Added /all to softblock

Pamasich:
With this PR, the softblock mod actually hides threads on /all too.

### DIFF
--- a/mods/softblock/softblock.json
+++ b/mods/softblock/softblock.json
@@ -1,7 +1,7 @@
 {
   "name": "Softblock magazines",
   "author": "shazbot",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "label": "Softblock magazines",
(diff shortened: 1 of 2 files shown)


aclist:
I might have found a few bugs in the mod itself while testing, trying to reproduce:

- Sometimes (probably) the cache of blocked mags does not get cleared correctly, and the mod is blocking mags that shouldn't be in the banned list anymore. This might be a collision with identical magazine names on different hostnames
- The mod is inserting the softblock icon into the sidebar (Subscribed/Random magazines widgets), but only for the last entry (see below)

a

aclist:
> Sometimes (probably) the cache of blocked mags does not get cleared correctly, and the mod is blocking mags that shouldn't be in the banned list anymore. This might be a collision with identical magazine names on different hostnames

Upon further investigation, it seems that the first time the script is installed (i.e., starting from nothing, no cache), the call to `softBlock()` on line 260 passes an undefined parameter, because the content of localStorage is empty. This causes the softblock icons to do nothing when clicked. 

After toggling the mod on and off, a dummy cache is created and the icons start working again.

Issue #269, opened 2024-04-05
[CHORE] 5.0.0 release


aclist:
- [ ] Fix concat funcs script to clear ESLint errors (dangling commas)
- [ ] Abstract fetch subscriptions function to shared utility functions
- [ ] Use fetch subs function in checkmarks mod
- [ ]  Abstract getComputedFontSize() function to safeGM
- [ ] Implement stricter CSS on noti panel (cf. #251)
- [ ] Expand troubleshooting section in docs with common issues
    - https://developer.chrome.com/docs/extensions/reference/api/userScripts#developer_mode_for_extension_users
- [ ] Add screenshots to links field
    - [ ] Document purpose of link/preview key in manifest

melroy89:
You might want to rename it to mbin/kes

aclist:
It seems reasonable. The name can be changed entirely if necessary. But I have not thought of a satisfactory alternative yet.

melroy89:
Yeah, I'm afraid /kbin is no more. Also feel free to contribute directly to the Mbin core. So your improvements will be available for everybody by default, without the need for additional CSS scripts. Assuming there are some CSS improvements that is beneficial for everybody. And even then, special features can be placed behind settings/switches. 


We are on Matrix: [https://matrix.to/#/#mbin-general:melroy.org](https://matrix.to/#/#mbin-general:melroy.org).

![image](https://github.com/user-attachments/assets/ba83f2c0-70a9-459e-b604-83b89ac2117d)

aclist:
This ticket is being closed and subdivided into atomic tickets.

PR #312, opened 2025-01-16
[FEAT] Simplified toggling the thread_checkmarks mod

Pamasich:
The thread_checkmarks mod was missing the depends_on (and depends_off for omni) property, so I added them both. Now the user shouldn't have to manually go look for the omnibar mod anymore.

### DIFF
--- a/mods/omni/omni.json
+++ b/mods/omni/omni.json
@@ -30,5 +30,8 @@
       ]
     }
   ],
-  "page": "general"
+  "page": "general",
+  "depends_off": [
+    "thread_checkmarks"
+  ]
 }
--- a/mods/thread_checkmarks/thread_checkmarks.json
+++ b/mods/thread_checkmarks/thread_checkmarks.json
@@ -3,7 +3,7 @@
   "author": "shazbot",
   "version": "0.1.0",
   "label": "Checkmark on subbed mags",
-  "desc": "Append a checkmark to magazines on the thread index you are subscribed to. The 'Subscriptions omnibar' option (page: General) must be enabled for this to work",
+  "desc": "Append a checkmark to magazines on the thread index you are subscribed to.",
   "login": false,
   "recurs": true,
   "entrypoint": "thread_checkmarks",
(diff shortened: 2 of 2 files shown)


aclist:
Nice fix, although I noticed two possible flaws in the MES design itself that bear consideration:

1. The user's checkmarks (i.e., the subscriptions) don't get updated on the fly unless they explicitly go to the omnibar mod's settings page and toggle it on and off to refresh cache. Since omnibar is an invisible dependency of checkmarks, this behavior might not be clear to the user.

2. In some rare cases, the checkmarks mod might get unintentionally turned off by the user:
- The user turns on the checkmarks mod
- This turns on the omni mod in the background
- The user later visits the omni mod's page and decides they don't want this mod on
- This turns off the checkmark mod in the background with no indication
- Later, the user wonders why the checkmark mod is not working

On the one hand, this is a flaw because the checkmarks mod could probably recycle the subscription fetch logic from the omni mod instead of listing it as a dependency.

On the other hand, maybe the flaw is just that neither mod explicitly states these dependencies upfront to the user (indeed, that is not a currently supported feature).

But in the latter case, one wonders if the user would be able to make meaningful use of this dependency information even if presented with it.

Pamasich:
I think this mod being dependent on Omnibar is a bit weird from a user perspective anyway. Like, it's not like the actual omnibar is used at all. It's just about getting the list of subscriptions. It makes sense to not reimplement everything in every mod, but we should probably rather move that logic out of omnibar to MES itself as a helper function, so both Omnibar and this one can access it.

aclist:
> I think this mod being dependent on Omnibar is a bit weird from a user perspective anyway. Like, it's not like the actual omnibar is used at all. It's just about getting the list of subscriptions. It makes sense to not reimplement everything in every mod, but we should probably rather move that logic out of omnibar to MES itself as a helper function, so both Omnibar and this one can access it.

Yes, agreed.

Pamasich:
Ah, I didn't realize that renaming the branch would "delete" it. It was named after the wrong mod.

aclist:
Alright. Interestingly enough, this idea was already floated ca. 8 months ago here: https://github.com/aclist/kbin-kes/issues/269

PR #329, opened 2025-01-20
chore: drop kbin style from collapsible comments

aclist:


aclist:
@Pamasich while making this fix, I noticed that mbin carries over internal theme parameters from kbin and doesn't bother to rename them to `mbin`:

kbin

. But interspersed among these are also some new `--mbin` parameters that are just wrappers around the old `--kbin` variant:

mbin

This means that the theme abstraction used in this mod and others (i.e,. inheriting colors from mbin's own theming settings, rather than hardcoding hex colors) continues to work, but if these parameter names are changed in the future, it might break the visual motif of some mods.

PR #335, opened 2025-01-22
refactor: Moved loadMags out of the omnibar mod

Pamasich:
This PR is a continuation of #312.

I reimplemented the user magazine loading functionality of the omnibar mod in the safegm file instead and modified both omnibar and thread_checkmarks to use that shared function. This also allowed me to remove the dependency of the latter on the former entirely.

The default magazine loading logic for logged out users is still in the omnibar mod.

Btw, I probably won't be able to work on MES for the remainder of the week.


aclist:
> Btw, I probably won't be able to work on MES for the remainder of the week.

Noted.

Pamasich:
@aclist If I wanted to move the magazine fetching logic to a separate file (because it doesn't really belong in safeGM in my opinion, I just put it there because the others were all there), does the current build logic support arbitrary files in the helpers directory out of the box, or do I need to take any additional steps aside of creating the file?

aclist:
In theory, could change the build script to just ingest all JS files from a special JS-only directory, but I preferred to be explicit with the additional dependencies here rather than globbing the whole contents of a directory without checking the files. I'm open to changing that, though.

Pamasich:
Hmm, I was planning to switch this to a draft PR so it would be clear that I'm intending to make changes to this before it gets merged, but I can't seem to find a button for that here. So this comment will have to do.

aclist:
I've started work locally on a PR that will store JS helper files in their own directory (`helpers/js`) and refactors the corresponding build scripts and MES main file to support this logic. It should allow for dropping helper JS files into that area without having to manually declare them in any build scripts. This currently affects `safegm.user.js`, `funcs.js`, and `pages.js`.

Also on that note:

- I don't know why safeGM uses the `user.js` notation; that should be dropped
- Following this patch, functions deemed unnecessary for the context of safeGM can be easily moved out into their own file(s) in the same directory

I haven't checked in any code yet.

Pamasich:
One more recently introduced bug and some cosmetic changes remaining until this PR is finished. I probably can't work on it today, but my goal is to have it finished this week.

aclist:
Thanks. If you pull beta 37 and resolve the merge conflicts in the mod, you should be able to take advantage of the bugfixes to #465 (the whole keytrap kludge was ripped out), #466, and be able to pop/unpop a loading dialog using #470.

Pamasich:
I tried adding the loading dialog, but it doesn't seem to show up. I'm currently looking into the issue.

Pamasich:
Oh, I didn't realize I have to add it to the document manually. Now it works.

Pamasich:
I'm pretty much finished here. I'll give both mods another quick check tomorrow to make sure they work and then ready this PR. I'll summarize what I changed in a comment.

Pamasich:
I've finally finished implementing this. Here's a quick overview of what changes with this PR:

### loadMags

The functionality of loading the user's subscriptions has been moved to this function in the `safeGM` file. It first tries to load the user's subscriptions from the sidebar, then the profile if necessary. It takes a callback which it runs after successfully fetching magazines via either attempt.

Besides the callback, loadMags takes three more parameters currently:
- ns: a namespace/id used for the safeGM setValue key that allows cancelling the callback later with `loadMags.cancel()` in a mod's teardown logic.
- useCache: optional boolean parameter which determines whether the cached result is used. The result is cached regardless. Caching requires the mod to implement some way to refresh the cache, be it with a button or by clearing the cache on teardown.
- runCallbackOnlyOnce: optional boolean parameter which can be used by mods that aren't optimized for running their setup code twice, like if it constructs a UI (and then replaces it the second time the callback gets called, even though the user was using it).

The function does its own check for whether the user is logged in, as I can't control from where it is called.

### thread_checkmarks

This mod has been changed to not use caching by default, but it provides it as an option in case the user doesn't have the sidebar subs enabled. By not using caching, users don't have to turn the mod off and on again anymore. And if subs are in the sidebar, especially if sorted by recent activity, most if not all of the checkmarks should be instantaneously visible anyway.

### omni

This mod is the reason I had to include the option to only run the callback once. The goal of running it twice is to provide some results early and add the rest when everything has finished loading. But this mod constructs a whole UI, and replaces it when the callback is called again. That kind of betrays the purpose of running it twice in the first place, as the user would be using the early results only to have the UI close out of nowhere while they're using it.

So with this mod, I've kept it working the same as before, with the addition of the loading indicator dialog. The loading indicator won't be shown automatically, it behaves the same as the actual omnibar (shown and hidden via metakey) and gets automatically replaced with the omnibar if the callback finishes while it's visible.

mods/omni/omni.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "Subs omnibar",
   "author": "shazbot",
-  "version": "0.1.1",
+  "version": "0.2.0",
   "label": "Subscriptions omnibar",
   "desc": "Quickly jump to a magazine by viewing your subs in a floating search panel (type to search). Use the meta key to open/close the omnibar. The meta key will not take effect inside of text fields. If you are logged out, the omnibar shows default magazines (most popular). To refresh the cached list of your magazines or change settings while this add-on is enabled, toggle off and then on again.",
```
aclist: While it's true that toggling the mod is used to clear cache, I don't think it's actually true anymore that you have to toggle the mod to change your settings. You can update the meta key and add the tapbar dynamically without toggling the mod.

helpers/safegm.user.js
```diff
@@ -234,6 +234,110 @@ function getPageType () { //eslint-disable-line no-unused-vars
...
+ * to provide some early results. This can be skipped by using the runCallbackOnlyOnce
+ * parameter.
+ * @param {function(string[]):void} callback The parameter contains the full names (with instances)
+ * of the user's subscriptions
+ * @param {string} ns A namespace used for the cancellation key (ideally the mod's name)
+ * @param {boolean} useCache Whether the cached result should be used
+ * @param {boolean} runCallbackOnlyOnce Workaround for mods that aren't optimized for running the
+ * callback twice. When true, the callback is only executed once when all mods are loaded.
+ */
+async function loadMags (callback, ns, useCache, runCallbackOnlyOnce) {
```
aclist: Since the `runCallBackOnlyOnce` parameter is somewhat of a specialized use case that may not be apparently clear, maybe it would be good to explicitly set the default parameter to false? `async function loadMags(callback, ns, useCache, runCallbackOnlyOnce=false)`

aclist:
>  This mod has been changed to not use caching by default, but it provides it as an option in case the user doesn't have the sidebar subs enabled. By not using caching, users don't have to turn the mod off and on again anymore. And if subs are in the sidebar, especially if sorted by recent activity, most if not all of the checkmarks should be instantaneously visible anyway.

One thing I wanted to double-check, as I'm not sure if it's intended/desirable behavior:

Using thread checkmarks:
- User has no subscriptions sidebar
- User loads the page and enables the mod in MES settings, cache is initially disabled
- Toggle triggers the mod, which calls `loadMags()`, fetches the remote mags, and then calls `runCallback()`
- User then ticks the `Caching: Enable` box, which updates the setting
- **Ticking this box also triggers MES to run the mod again**, which calls `runCallback()` and then aborts

MES stepping through the mod again is just a side effect of its design, and the callback will just abort its behavior if it finds that the changes to the page were already applied (e.g., checkmarks already exist).

But what if we follow up with this scenario?

- Continuing from the above, with the mod still on, the user now unticks the `Caching: Enable` box
- **This triggers MES to run the mod again, and since caching is off, it tries to fetch the mags from the remote again**
- But the mod was already on and the mags were already fetched...once again, the elements are already applied on the page, so there will be no cosmetic change, but now an async function is trying to pull subscriptions again in the background

Anything we can do to improve upon this scenario from a performance perspective?

Also, considering the above, is it actually the case that you need to toggle the mod on and off to reset cache? Couldn't you just tick/untick the checkbox?

mods/omni/omni.user.js
```diff
@@ -420,21 +360,50 @@ function omniInit (toggle) { // eslint-disable-line no-unused-vars
...
+                $(load).hide();
+            } else {
+                $(load).show();
+            }
         }
     }
 
     if (toggle) {
-        $(document).off("keypress.omnikey");
+        const modal = makeLoader(id, "Please wait for the subscriptions to finish loading...");
```
aclist: Do you mind if we change this to more of a statement of fact than a request? I'm going to update `makeLoader()` to always prefix the incoming loading message with `MES: ` so that it's clear who owns this dialog (not Mbin), and I feel like it's kind of implied by the loading dialog that you need to wait for something to load, so I think messages should be pithy and state "what" the mod is currently doing (fetching, filtering parsing), rather than what the user needs to do ("please wait"). E.g.,

"Please wait for the subscriptions to finish loading..." -> "MES: Fetching subscriptions list..."

helpers/safegm.user.js
Pamasich: Sure, I'll also add it to `useCache` to make it more obvious that it's optional too.

mods/omni/omni.user.js
Pamasich: Sure, that works too.

aclist:
> Anything we can do to improve upon this scenario from a performance perspective?

One thing that would be totally feasible to do is start sending an additional parameter into the triggered mods to tell them that this was a toggle event and not a regular occurrence. The mod could process this incoming flag and take some action accordingly.

Pamasich:
> But the mod was already on and the mags were already fetched...once again, the elements are already applied on the page, so there will be no cosmetic change, but now an async function is trying to pull subscriptions again in the background

That is the consequence of bypassing the caching, yeah. I don't see it being too problematic with a caching toggle you'll trigger maybe once or twice, but I can see the issue with other options that might be used more frequently while experimenting with mods. It's another reason to provide a choice between caching or not caching I guess. Heaving it turned off makes things simpler, but at a cost.

> One thing that would be totally feasible to do is start sending an additional parameter into the triggered mods to tell them that this was a toggle event and not a regular occurrence. The mod could process this incoming flag and take some action accordingly.

I've been thinking in general that it would be good to know how a mod's entry function was triggered (page load, mod toggle, a minor setting change, or a mutation), and in the case of a setting, which one. That way we could for example have the `thread_checkmarks` mod just change the color of existing checkmarks when a new color is chosen, instead of running the entire mod again on every color change.

> Also, considering the above, is it actually the case that you need to toggle the mod on and off to reset cache? Couldn't you just tick/untick the checkbox?

I haven't thought of that, good idea. I'll try implementing that instead.

Pamasich:
Like, right now, toggling the cache checkbox just changes the value passed to loadMags. And resetting the cache is handled in `thread_checkmarks`, so currently it's not triggered by unticking the checkbox.

Pamasich:
Wait no, I can't implement it currently. I would have to know that this specific setting just got changed, or I'd have to clear the cache every time the mod is run with caching turned off. Considering all mods currently use the same list, that would be pretty bad.

aclist:
> Wait no, I can't implement it currently. I would have to know that this specific setting just got changed, or I'd have to clear the cache every time the mod is run with caching turned off. Considering all mods currently use the same list, that would be pretty bad.

Alright, it is no problem, let's go with what you have at the present moment, which is already quite good.

> I've been thinking in general that it would be good to know how a mod's entry function was triggered (page load, mod toggle, a minor setting change, or a mutation), and in the case of a setting, which one. That way we could for example have the thread_checkmarks mod just change the color of existing checkmarks when a new color is chosen, instead of running the entire mod again on every color change.

Well, I think it'd be like 5 LOC and some enums to start sending the type of entry. The "any change triggers the whole mod to reload" design was, I think, a concession to simplicity, because it lets mods just have a very simple logic and no matter what the triggering behavior was, the mod will achieve the desired effect without complex checks or conditions. But parsing this extra parameter doesn't have to be mandatory; it could be something available for those mods that want it.

As for the type of setting changed specifically, every element in the settings menu already has a property which defines its originating key in the JSON, so the data is definitely there to pass around. It's eminently feasible.

aclist:
> Wait no, I can't implement it currently. I would have to know that this specific setting just got changed, or I'd have to clear the cache every time the mod is run with caching turned off. Considering all mods currently use the same list, that would be pretty bad.

It's pretty straightforward, so I sketched out a simple skeleton for this functionality in #496. I can probably get a basic prototype up and running tomorrow. In the context of this mod, I think you'd just test the event type against `Trigger.Setting` and check whether the metadata parameter matches the key `refresh`.

aclist:
Trigger functionality was merged into `testing` if you want to add the check for the cache: enabled button being clicked.

Pamasich:
I now made it so changing either of the settings doesn't run the usual code. Turning caching off clears the cache now (that action has been removed from turning the mod off), and changing the color does simply that.

aclist:
Your last commit has me thinking, does it make sense for MES to call a mod when:

- The mod is explicitly OFF
- A setting was just changed **while** OFF?

Shouldn't we just be storing the setting and doing nothing?

I can't think of a use case for why MES needs to call mods here, since it's not a teardown request, it's merely an internal settings change. This should result in a no-op and not put the extra setup burden on the mod.

Or did I overlook something? (Just came to the top of my head.)

Pamasich:
Yeah, it would probably be best to not call mods in that scenario. It was unexpected to see the cache still being cleared when I changed the setting while the mod was off.

aclist:
Alright, I've updated that behavior here: https://github.com/aclist/kbin-kes/commit/4c58e1e8b9f8e7f7c780572113eb283da66a136f

This doesn't break the behavior of your PR as such, so I am fine to approve it after testing it. If you want to update the logic at some point later on, it is fine by me. I know you've already made multiple changes to this PR already.

Pamasich:
I removed the commit which added the relevant check.

mods/omni/omni.user.js
```diff
@@ -420,21 +360,50 @@ function omniInit (toggle) { // eslint-disable-line no-unused-vars
...
-            });
+            if ($(`#${id}-filter-modal-bg`).is(":visible")) {
+                $('.kes-omni-modal').show();
+            }
+            clearLoader(id);
+        }
+    }
+
+    function kickoffListener (e) {
+        if (e.key !== code) return
```
aclist, line 371: Not a regression, but found this bug: I noticed that if you change the meta key multiple times while the mod is on, eventually the meta key will stop responding. It gets updated in localStorage, and the `kickoffListener` function, which is bound to the `keypress.omnikey` namespace attached to the document, evaluates whether the pressed key matches--but there's actually no cleanup action on the old namespaced listener (it only gets cleaned up on explicit toggle off actions), which contains the event listener for the meta keycode.

So it seems that multiple event listeners are getting bound to this old element, and it becomes a race condition to see if the key currently configured happens to hit the right listener in the stack. (When it doesn't, sometimes the old keycode will open the modal). Sometimes it triggers the modal, but sometimes not. Given that it's possible for the mod to be on and the meta key to be changed while that happens, it seems like we should be catching the `Trigger.Setting`, remove the old listener, and reinsert a clean listener anew.

This was an oversight in the last refactor I did, so I've pushed a commit into this PR that seeks to resolve this behavior.

helpers/safegm.user.js
```diff
@@ -228,6 +228,110 @@ function getPageType () { //eslint-disable-line no-unused-vars
...
+        genericXMLRequest(url, (response) => {
+            const dom = new DOMParser().parseFromString(response.responseText, "text/html");
+            // get the magazines from this page
+            mags.push(
+                ...Array.from(dom.querySelectorAll('#content .stretched-link'))
+                    .map((link) => link.getAttribute('href').split('/')[2])
+            );
+            // load more pages if there are
+            const nextPage = dom.querySelector('#content .pagination__item--next-page');
+            if (nextPage.hasAttribute('href') && nextPage.href != window.location.href) {
```
aclist: I ran into a blocking `Uncaught TypeError: nextPage is null` that can occur if the user has no magazine subscriptions, no subscriptions sidebar enabled, and therefore no next page button.

aclist (changes requested):
Following the fix I just pushed, should only need to fix the `nextPage` issue.

Pamasich:
@aclist I saw two issues with the changes you made which I fixed. One being that the mod wasn't triggered on page load. The other was that you didn't remove the line about removing the modal from the main function, meaning even on setting changes the modal was still destroyed, but due to the changes never recreated.

I also simplified the jsdoc comments now that the sphinx documentation is a thing. I mainly decided to go for jsdoc originally to give the parameters proper types (for intellisense), so I've reduced them to just that + a single line description. If someone wants a more detailed description, they can go to the proper documentation.

aclist (approved):

aclist:
Thanks for catching that. Everything looks good! It seems perfectly fine to have JSDoc syntax at the code level for use in an editor coexist with Sphinx at the docs level. I basically just moved the comments verbatim into the Sphinx side.

Issue #354, opened 2025-01-27
[CLARIFICATION REQUESTED] About the timestamp mod


Pamasich:
### Browser

Chromium-based

### Script manager

Tampermonkey

### KES version

4.3.0-beta.8

### Describe the feature

I'm a bit confused by the timestamp mod. Why does it reload the page when turned off? I was wondering if it's really necessary, but I can't even find the code that does it in the first place. The teardown seems to literally just be an empty return statement.

aclist:
See line 1239 of `kes.user.js`:

```
        //necessarily reload the page when verbose timestamps are toggled off
        //otherwise, triggers a loop of mutations because reverting timeago mutates watched node
        if ((func === "timestamp") && (state === false)) {
            window.location.reload();
        } else {
            toggleSettings(func);
        }
```

IIRC, the global MES MutationObserver triggers on changes to the page index as a means of watching for infinite scroll scenarios, when hiding some thread and repopulating it with something new, and when replying to a post (i.e., page has new content that mods need to be applied to). Resetting the timestamps triggers a change that causes infinite recursion, so an exclusion was added for this mod in particular.

Maybe there is a more satisfactory workaround.

Pamasich:
Hmm. But setting up the mod also seems to cause the same infinite recursion issue. I added `console.log(times)` and it keeps outputting the results over and over again.

aclist:
> Hmm. But setting up the mod also seems to cause the same infinite recursion issue. I added `console.log(times)` and it keeps outputting the results over and over again.

The mbin timestamp counter increments every second until one minute has elapsed. (This will display as: "Posted 1s ago...", "Posted 2s ago...", etc.) After a minute has passed, it stops updating as frequently. During this initial ramp-up the mutation observer should thus catch and emit changes each time, but allow them to silently pass until the 60 second mark.

Pamasich:
> The mbin timestamp counter increments every second until one minute has elapsed. (This will display as: "Posted 1s ago...", "Posted 2s ago...", etc.) After a minute has passed, it stops updating as frequently.

Yeah, I didn't think of that. Every time one of the relative time stamps gets updated, the mod gets run. That's definitely what I was seeing there.

Pamasich:
But then I'm not sure about the infinite recursion. I don't really see anything wrong, though I might not be looking in the right places.

[...]

PR #361, opened 2025-01-29
fix: timestamps

aclist:
Resolves multiple issues brought up in https://github.com/aclist/kbin-kes/issues/354:

- Mod required hard reload for teardown

The mod now makes a clone of the `timeago` elements and hides the original ones. On teardown, the clones are cleared and the original elements are made visible again. Their human readable timestamps continued updating in the background, so on teardown, the relative time delta is still accurate.

- Unnecessary recursion

Increments to the `timeago-id` parameter of `timeago` elements are silently ignored, and if a `timeago` element has already been cloned, the clone action is not propagated again.

- Update UTC/local time variant in place

If the user has changed their time setting preference while the mod is already on, the mod recursively updates existing timestamps to their corresponding variant (UTC<->local). The time variant is stored inside the dataset attribute `data-time-variant` and matches the corresponding value in the localStorage.

- Special exclusion inside of MES for this mod

Dropped requirement for MES to perform additional handling for the timestamps mod.

### DIFF
--- a/kes.user.js
+++ b/kes.user.js
@@ -1399,17 +1399,23 @@ function constructMenu (json, layoutArr, isNew) {
                 }
                 return
             }
+            //trigger when username popover dialog is spawned on hover
+            //there can only be one popover spawned at a given time
+            if (mutation.target.id === "popover"){
+                applySettings("timestamp");
+                return
+            }
+            //workaround for timeago ticks changing timestamp textContent
+            //implies that the active 60s timestamp is updating
+            //see also updateState()
             if (mutation.target.className === 'timeago') {
-                //workaround for timeago ticks changing timestamp textContent
-                //implies that the active 60s timestamp is updating
-                //reapplies verbose timestamps
-                //see also updateState()
                 if (mutation.target.textContent.indexOf("ago") >= 0) {
                     applySettings("timestamp");
                 }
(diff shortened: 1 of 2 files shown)


Pamasich (approved):
Looks good to me. I've tested it a bit and haven't noticed any bugs.

Pamasich:
I did notice though that the mod doesn't yet apply to the user popup (when hovering over a username in the thread index). Not really in scope for this fix, but seems like there's still more to fix here.

aclist:
If you give it a few seconds, the popup will actually update. (Not intended, but it works). I'm not sure why there is a delay, possibly because it takes a moment for the DOM to update after the popup action is triggered.

aclist:
Disregard the above. I am making some changes to make it apply more consistently to the popover.

Pamasich:
> If you give it a few seconds, the popup will actually update. (Not intended, but it works). I'm not sure why there is a delay, possibly because it takes a moment for the DOM to update after the popup action is triggered.

That sounds to me like it's probably updating when the mod gets executed again because one of the relative timestamps got changed by Mbin.

I've given it a quick test, gave it some minutes even, and haven't seen it update. Probably because all the timestamps on the page were over an hour old, so they took ages to get updated.

aclist:
Popover issue should be fixed now.

PR #371, opened 2025-01-31
feat: get page type

aclist:
This is a basic prototype for a helper function that returns the page type by splitting the `window.location.href`. Because there are numerous subconditions for URLs that need to be tested (e.g., favorites page, but inside of a thread) , a simple enum (frozen object) was not used.

Mods can use this to simplify taking/not taking some action on specific pages.

This should cover most scenarios, but there are likely some page types missing because I simply don't know about them.

If no valid page type could be found, returns the string `Unknown`.


aclist:
I think it's possible for the following sections to have pagination if there are too many entries, so the match logic should be more expansive (`includes` rather than `===`):

- favorites
- boosts
- followers
- following
- user comments/posts/replies/threads

PR #382, opened 2025-02-12
fix: abort login on notipanel

### LINKED ISSUE #380: [BUG] Notifications panel needs better abort logic if user is not logged in
### Browser

Firefox-based

### Script manager

Tampermonkey

### KES version

4.3.0-beta.12

### Describe the bug

This mod expects various elements to be onscreen but does not abort gracefully if they are missing. This causes junk to be emitted to the console if the user is on the right page but is not actually logged in to the site. There is already a `isLoggedIn()` helper function that can be used here, so it should be added before any other action is taken. Currently the script is trying to build the panel on the assumption that the login element will be found.

aclist:
Resolves #380


aclist:
This PR can be rescinded once the PR discussed in #386 is merged.

aclist:
This PR was made obsolete by #408; rescinding

PR #385, opened 2025-02-13
feat(hide_sidebar): expand content when sidebar is hidden

Pamasich:
When the entire sidebar is hidden, the space it occupied is still left empty. This PR adds an option to have the main content expand to make use of the freed up space.


aclist:
Thanks. I'll try to make a dent in this and other tickets over the weekend. Currently got the flu and I am progressing at a snail's pace.

aclist:
Nice fix, I liked it.

PR #446, opened 2025-02-18
fix: mag instance names

### LINKED ISSUE #398: [BUG] mag_instance_names doesn't apply to magazine indexes
### Browser

Chromium-based

### Script manager

Tampermonkey

### KES version

4.3.0-beta.15

### Describe the bug

The mod doesn't currently apply to the magazines page or the magazine list in the sidebar.

aclist:
Resolves #398

### DIFF
--- a/mods/mag_instance_names/mag_instance_names.user.js
+++ b/mods/mag_instance_names/mag_instance_names.user.js
@@ -1,40 +1,51 @@
 function magInstanceEntry (toggle) { // eslint-disable-line no-unused-vars
-    // ==UserScript==
-    // @name         Magazine Instance Names
-    // @namespace    https://github.com/aclist
-    // @version      0.1
-    // @description  Shows instance names next to non-local magazines
-    // @author       artillect
-    // @match        https://kbin.social/*
-    // @license      MIT
-    // ==/UserScript==
-    const path = window.location.href.split('/')
-    if ((path[3] === "m") || (path[3] === "magazines")) return
-    function showMagInstances () {
-        $('.magazine-inline').each(function () {
-            // Check if community is local
-            if (!$(this).hasClass('instance')) {
-                $(this).addClass('instance');
-                // Get community's instance from their profile link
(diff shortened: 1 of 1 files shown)


Pamasich:
There is a bug with this, the magazine name isn't properly cleaned up if the magazine doesn't have an icon.

image

image

image

I'm only observing this issue in the subscriptions in the sidebar

Pamasich:
Also, cleanup in general doesn't seem to be fully clean:

**Before enabling the mod**
image

**After disabling the mod**
image

Seems like there's a `display: block;` that doesn't get cleaned removed.
image

aclist:
I didn't have icons enabled. I wasn't able to reproduce the first part.  Fixed the second part. I subscribed to those same magazines, and the name was cleaned up.

Pamasich:
> I wasn't able to reproduce the first part.

Interesting. I wonder then if this is something the browser is doing.

What's happening for me is that the `` tag, which contains the icon if it exists, is turned into a `<span>` tag instead if there's no icon. Since the selector simply looks for the first `<span>` tag, it thus grabs the "icon" instead of the name container it was looking for.

Example:
```html
<a class="stretched-link" href="/m/climate@slrpnk.net">
    <span class="magazine-subscription-avatar-placeholder"></span>
    <span class="magazine-name">
         <span>Climate -  truthful information about climate, related activism and politics.</span>
    </span>
</a>
```

Even if I disable the icons, the HTML still looks like this for me. It simply adds a `mobileOnly` class to both spans if icons are disabled. I've also tried changing other settings, with no change to the HTML. So I wonder if this is something the browser does, for optimization maybe.

A fix should be simple to implement. Just check for the `magazine-name` class instead.

aclist:
I see. Even with MES off, that first span tag gets set to `display: none` from some inherited property for me. I thought thay maybe some Mbin setting is disabling empty avatars, but I looked through the sidebar and enabled everything that seemed relevant.

In any case, it's as you say, just need a more granular selector.

aclist:
Should be fixed now.

PR #460, opened 2025-02-20
fix (federation awareness): borders on microblogs

### LINKED ISSUE #397: [BUG] Federation Awareness not working on Microblog
### Browser

Chromium-based

### Script manager

Tampermonkey

### KES version

4.3.0-beta.15

### Describe the bug

As the title says, the federation awareness mod currently has no effect on the Microblog view. I don't know if that's by design or a bug, but it seems weird to me as a user.

aclist:
Resolves #397. Contains a number of fixes:

- Apply bubble and border styles correctly on microblogs
- Make the bubble box shadow ("fuzz") effect optional
- Enforce applying borders on the right, rather than right, left, or both, because the latter two options were never applicable on comments
- Make scale slider apply to both bubbles and borders uniformly
- Reduce the range of the slider scale (it was possible to create elements that were abnormally small or large)
- Apply different scale formula to borders and bubbles
- Use real hex color provided by colorpicker instead of derived value

### DIFF
--- a/mods/kbin_federation_awareness/kbin_federation_awareness.user.js
+++ b/mods/kbin_federation_awareness/kbin_federation_awareness.user.js
@@ -34,7 +34,7 @@ function initKFA (toggle) { // eslint-disable-line no-unused-vars
                 header div.data-moderated,
                 header div.data-home,
                 article .data-federated,
-                article .data-moderated
+                article .data-moderated,
                 article .data-home {
                     display: inline-block;
                     width: ${scale}px;
@@ -97,6 +97,7 @@ function initKFA (toggle) { // eslint-disable-line no-unused-vars
 
     function kfaStartup () {
         kfaInitClasses();
+        safeGM("removeStyle","kfaInjectedCss");
         safeGM("addStyle",kfaGenCSS(),"kfaInjectedCss");
     }
 
@@ -213,7 +214,7 @@ function initKFA (toggle) { // eslint-disable-line no-unused-vars
     }
 
     if (toggle) {
-        kfaShutdown();
+        //kfaShutdown();
         kfaStartup();
     } else {
         kfaShutdown();
--- a/mods/kbin_federation_awareness/kbin_federation_awareness.user.js
+++ b/mods/kbin_federation_awareness/kbin_federation_awareness.user.js
@@ -1,8 +1,6 @@
 function initKFA (toggle) { // eslint-disable-line no-unused-vars
-    /*
-        License: MIT
-        Original Author: CodingAndCoffee (https://kbin.social/u/CodingAndCoffee)
-        */
+    //License: MIT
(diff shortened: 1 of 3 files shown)


Pamasich:
The border seems only applied to the replies, not the post itself. For Microblogs.

image

Pamasich:
It's because the selector being used doesn't cover them:

image

image

mods/kbin_federation_awareness/kbin_federation_awareness.user.js
```diff
@@ -13,145 +11,110 @@ function initKFA (toggle) { // eslint-disable-line no-unused-vars
...
-                box-shadow: 0 0 3px 2px ${fedColor0};
-                background-color: ${fedColor0};
-                margin-right: 4px;
-                margin-left: 4px
+        }
+        if (style === "border") {
+            const scale = setScale(indicatorScale, 10)
+            const borderCSS = `
+            article.data-federated,
+            .comment.data-federated {
```
Pamasich: Add `.post.data-federated,`

mods/kbin_federation_awareness/kbin_federation_awareness.user.js
```diff
@@ -13,145 +11,110 @@ function initKFA (toggle) { // eslint-disable-line no-unused-vars
...
-                display: inline-block;
-                width: ${setScale}px;
-                height: ${setScale}px;
-                border-radius: 10px;
-                box-shadow: 0 0 3px 2px ${modColor0};
-                background-color: ${modColor0};
-                margin-right: 4px;
-                margin-left: 4px;
+            article.data-moderated,
+            .comment.data-moderated {
```
Pamasich: Add `.post.data-moderated,`

mods/kbin_federation_awareness/kbin_federation_awareness.user.js
```diff
@@ -13,145 +11,110 @@ function initKFA (toggle) { // eslint-disable-line no-unused-vars
...
-                display: inline-block;
-                width: ${setScale}px;
-                height: ${setScale}px;
-                border-radius: 10px;
-                box-shadow: 0 0 3px 2px ${homeColor0};
-                background-color: ${homeColor0};
-                margin-right: 4px;
-                margin-left: 4px;
+            article.data-home,
+            .comment.data-home {
```
Pamasich: Add `.post.data-home,`

Pamasich (changes requested):

Pamasich:
I'm also noticing that expanding the replies under a post (from the initial 2) doesn't apply the border to the new replies.

image
*Image has the fix for the previous issue already applied, that's why the border is there on the post*

image

Pamasich:
This one looks to me like an issue with the replies possibly not triggering the mutation listener though, not a mod-specific issue.

aclist:
Should be working now on both fronts (OP styling and microblog expansion).

PR #473, opened 2025-02-26
fix: namespaced omni keypress listener

### LINKED ISSUE #466: [BUG] (omnibar) Meta key still works after turning the mod off
### Browser

Chromium-based

### Script manager

Tampermonkey

### KES version

4.3.0-beta.35

### Describe the bug

When I turn the mod off, pressing the meta key still seems to work. It just throws an error since the modal doesn't exist anymore.

aclist:
Resolves #466


Pamasich:
`bind` and `unbind` are deprecated in favor of `on` and `off` btw.

PR #487, opened 2025-03-03
feat: cosmetic separator

aclist:
The top post in a thread and microblog does not follow the tiled gap behavior of other panels onscreen. This fix inserts an extra divider between the top post and navbar to simulate the right spacing.

### DIFF
--- a/mods/thread_separator/thread_separator.js
+++ b/mods/thread_separator/thread_separator.js
@@ -1,0 +1,15 @@
+function dividerInit (toggle) { //eslint-disable-line no-unused-vars
+    function insertSeparator () {
+        const pt = getPageType()
+        if (!isThread() && pt !== Mbin.Microblog) return
+        if (document.querySelector("#mes-thread-divider")) return
+        const top = document.querySelector(".section--top")
+        const sep = document.createElement("div")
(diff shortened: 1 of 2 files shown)


Pamasich (approved):

mods/thread_separator/thread_separator.js
```diff
@@ -0,0 +1,15 @@
+function dividerInit (toggle) { //eslint-disable-line no-unused-vars
+    function insertSeparator () {
+        const pt = getPageType()
+        if (!isThread() && pt !== Mbin.Microblog) return
+        if (document.querySelector("#mes-thread-divider")) return
+        const top = document.querySelector(".section--top")
+        const sep = document.createElement("div")
+        top.insertAdjacentElement("beforebegin", sep)
+        sep.style.height = "11px"
```
Pamasich: Right now the divider causes the post to be a bit further down than the sidebar for me:

image

I'd personally line them up by using the same value for the divider height as the sidebar uses for its padding there (.5rem).

image
*This image has the suggested change applied*

aclist: Thanks, I think I just eyeballed it at first and it seemed satisfactory, but I felt something was intuitively wrong.

PR #489, opened 2025-03-03
fix (rearrange): occlusion issues

### LINKED ISSUE #488: [BUG] (rearrange) "More" dropdown in opening post is occluded by activity bar
### Browser

Firefox-based

### Script manager

Tampermonkey

### KES version

4.3.0-beta.52

### Describe the bug

By default (before MES is on), the `#activity` element has an unused `z-index` property set to `5` that is suppressed by other stylesheets. When the mod is enabled and the order of elements is rearranged, this `z-index` property becomes active. If the activity bar is moved directly below the opening post, when hovering the "More" dropdown in the post options, the dropdown appears below the activity bar.

The z-index needs to be unset or normalized here when the mod is enabled.

aclist:
Resolves #488

### DIFF
--- a/mods/rearrange/rearrange.user.js
+++ b/mods/rearrange/rearrange.user.js
@@ -13,6 +13,9 @@ function rearrangeInit (toggle) { // eslint-disable-line no-unused-vars
 
         op.style.order = settings["op"]
         activity.style.order = settings["activity"]
+        //fix for #488
+        activity.style.zIndex = 0
+
         if (isLoggedIn()) {
             const post = document.querySelector('#comment-add');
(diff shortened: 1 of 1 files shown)


Pamasich:
I noticed the filter bar has the same issue and isn't fixed by this:

image

That image is from before applying the changes from this PR. With the PR, the activity bar is fixed but the filter bar still overlaps the menu.

Pamasich:
My settings:

image

aclist:
Thanks for catching that, I thought I tested them all correctly.

Pamasich (approved):

Issue #454, opened 2025-02-19
[CHORE] Update documentation to 5.0.0 spec


aclist:
### Browser

Firefox-based

### Script manager

Tampermonkey

### KES version

4.3.0-beta.33

### Describe the bug

Catch-all stub for documentation changes

aclist:
Resolved by #503

PR #490, opened 2025-03-03
chore: add code of conduct and contributing docs

aclist:
`CONTRIBUTING.md` has links currently pointing to the old documentation, but this will have to be updated following #454

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,32 @@
+Thank you for your interest in contributing to this project. Below are some basic guidelines and instructions.
+
+# Mods (scripts)
+Mods should focus on atomic features and do one thing well. If there is a need for multiple options within a mod, expose those via the JSON manifest to create granular settings.
+
+Mods should do not call other mods. If you are finding a need to call another mod from within a mod, this likely indicates a need for splitting that functionality into a public helper function 
+accessible to all mods.
+
+Mods are invoked in serial fashion by MES and their logic is walked through. Mods should not modify the main MES modal or MES itself.
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,29 @@
+Thank you for your interest in contributing to this project. Below are some basic guidelines and instructions.
+
+# Mods (scripts)
+Mods should focus on atomic features and do one thing well. If there is a need for multiple options within a mod, expose those via the JSON manifest to create granular settings.
+
+Mods should do not call other mods. If you are finding a need to call another mod from within a mod, this likely indicates a need for splitting that functionality into a public helper function accessible to all mods.
+
+Mods are invoked in serial fashion by MES and their logic is walked through. Mods should not modify the main MES modal or MES itself.
```
Pamasich: > Mods are invoked in serial fashion by MES and their logic is walked through.

What's the purpose of this sentence, in terms of its inclusion in the contributing.md? It doesn't seem to provide context for the succeeding sentence, nor is it a guideline/advice like the other lines.

Also, wouldn't it make sense to also include a section stating that mods need to provide a teardown implementation which restores the document to how it was before the mod was applied?

aclist: Honestly, no purpose. Also, maybe this whole section should be dropped in favor of pointing to a link to the real documentation. I wrote this part before updating the docs, but everything here is either covered already or in more detail in the docs.

aclist: Resolved per commits on 2025-06-19

PR #494, opened 2025-03-05
fix (code highlighting): mod not applying, buttons not working

### LINKED ISSUE #388: [BUG] code_highlighting doesn't do anything
### Browser

Chromium-based

### Script manager

Tampermonkey

### KES version

4.3.0-beta.15

### Describe the bug

It seems that `code_highlighting` isn't doing anything for me on Mbin 1.8.0. Haven't yet looked into why.

aclist:
Resolves #388

After dropping the dependency, the mod still exhibited problems with the event listeners attached to icons doing nothing. This patch fixes that behavior.


aclist:
Depends on #499  in order to trigger on nested reply mutation events, which are outside of the normal tree

aclist:
This should be finished now and supports the following cases:

- Apply on Trigger.Pageload
- Apply on Trigger.Mutation (replying to a comment by posting a codeblock, infinite scroll on comments)
- Clear mod on toggle off
- Update code highlighting themes in place (Trigger.Setting)
- Copy/collapse/expand buttons in header work

Pamasich:
I created a codeblock to test, and the code seems to be offset for me.

image
*How it looks with the mod on*

image
*What I typed*

image
*How it looks with the mod off*

Removing the code language doesn't resolve it.

Edit: seems to be because a `margin-left: 20px;` style is applied to all `.hljs-keyword` elements, but not the brackets. So I'm not sure if this is a new problem or if it simply hasn't been noticed so far.

aclist:
> Edit: seems to be because a margin-left: 20px; style is applied to all .hljs-keyword elements, but not the brackets. So I'm not sure if this is a new problem or if it simply hasn't been noticed so far.

I can't really see a reason for it, seems like it was just a cosmetic choice that was left in the CSS.

aclist:
It looks like it was being done in order to add padding to the header, but it was erroneously applying to the code block as well.

Pamasich (approved):

PR #498, opened 2025-03-06
fix: default_sort wasn't working

### LINKED ISSUE #389: [BUG] default_sort not working
### Browser

Chromium-based

### Script manager

Tampermonkey

### KES version

4.3.0-beta.15

### Describe the bug

The mod `default_sort` isn't currently working for me on 1.8.0. Have yet to look into why.

I didn't test the mod for the last version like almost all others, so I'm not sure if this is a 1.8.0 bug or if it existed prior.
Technically the mod is now obsolete, but it still has some relevance by being more granular than the native options and by supporting the magazine index page. So I still aim to get it to work for 1.8.0.

Pamasich:
Resolves #389 

This mod was originally written under the assumption, that one of the sort options on the page links to an unsorted url. The mod would then rewrite that option's url to use an explicit sort endpoint like the others. There only being explicitly sorted urls was taken to mean this mod has already been run, and so it aborted.
At some point it seems Mbin changed this so all options are always explicitly sorted. This means the mod now always assumes it's already applied, even when it isn't, and aborts itself even before redirecting the user.

This PR rips out all the code that's used to change the URLs the sort options point to, as that code is obsolete now anyway.


aclist:
I think we've talked about this before, but just to refresh my memory:

- If you navigate inside a thread in the default state, with no sort filter explicitly applied
- And you change the sort method for comments to something other than the default (and optionally turn the mod on, if it wasn't)
- The page redirects to the same thread with the `/top` suffix
- But once you are on anything other than the default, you can't change the suffix automatically without reloading the page
- Similar outcome on top-level pages like thread index and their sort order

So is the fact that the redirect happens on the default URL path, thereby letting you change the sort order once automatically, an unintended side effect? Or is the fact that you can't automatically change it more than once a bug?

Also, would the new `Trigger.Setting` functionality help here, in terms of catching these events?

I know that the stated purpose of the mod is "redirect from the default sorted page", so if the automatic behavior is not intentional, maybe it would be better to prevent changing the sort method once already on the page from triggering anything?

Pamasich:
I'm planning a general rework of the mod in the future anyway, the goal here was just to fix the bug so it's not broken anymore. I'll be looking some more into the actual functionality and if any trigger can be used once I get to that. I also planned to simplify the description and comments then.

When you land on a page like `kbin.earth/all` that isn't explicitly sorted, Mbin just applies a default sort to it. The goal here was to let the user override that, by redirecting them to their chosen option when they visit a page that isn't explicitly sorted.
So user visits `kbin.earth/all`, they get redirected to `kbin.earth/all/threads/newest` for example.

If you're on a page with an explicit sort option in the URL, I assume you intend to be there. That's why no redirect happens then.
The redirect from the non-sorted page is intentional however.

I do think it's not the best UX to redirect the user while the MES menu is open. When I get around to reworking the mod, I plan to make it so that the redirect only happens on page load. But that'll be a new PR.

aclist:
Alright, sounds good to me.

PR #501, opened 2025-03-07
feat (improved_collapsible_comments): Support microblog pages

### LINKED ISSUE #396: [FEAT] Implement collapsible replies on the microblog tab too
### Browser

Chromium-based

### Script manager

Tampermonkey

### KES version

4.3.0-beta.15

### Describe the feature

The `improved_collapsible_comments` mod doesn't cureently apply to the Microblog view. Neither does Mbin have a native collapsing mechanic there. (besides toggling ALL replies to a post on or off)

aclist:
Resolves #396

This PR is branched from #499


Pamasich:
I noticed an issue with teardown here. If you expand a post's replies with this mod enabled beforehand, and then turn the mod off, it's not properly reverted.
Teardown does work correctly if the mod is turned on while the post's replies are already expanded.

image
*How it should look*

image
*How it looks with the mod turned on*

image
*How it looks after turning the mod off*

Pamasich:
> Teardown does work correctly if the mod is turned on while the post's replies are already expanded.

I tried comparing this to the case that doesn't work, and found out I can reproduce the issue in this scenario by turning the mod on, tricking it into thinking it's off, and then turning it on again, then off.

So it seems to me like there's issues when the mod is run multiple times before teardown.

aclist:
It's probable that multiple stylesheets are getting added under the same id, and only one is getting peeled off. The mod needs to be refactored to support the trigger events better.

aclist:
Yeah, it was as simple as not applying the stylesheet again on subsequent mutations, since it already exists after pageload.

aclist:
> since it already exists after pageload

Edit: rewording...assuming the stylesheet was already applied by the time a mutation occurs is a bit risky, because the mod might have failed to apply when it was turned on (e.g. no Internet connection). So it would be better to actually query the document head for a stylesheet matching the right id before aborting.

Also: add some logic to safeGM to not let it insert multiple stylesheets under the same ID.

Pamasich (approved):

Issue #502, opened 2025-03-09
[CHORE] Dropping deprecated keys/methods


aclist:
### Browser

Firefox-based

### Script manager

Tampermonkey

### KES version

4.3.0-beta.56

### Describe the feature

While preparing the 5.x.x documentaion, I noticed some areas that are obsolete or could be improved upon:

`entrypoint` key
---
Mods used to be required to manually set up the name of their entrypoint function in the JSON manifest with this key. This is no longer relevant, as the build scripts automatically insert the key in the global manifest based on the root name of the mod. Even if a mod author adds something for the entrypoint key here, it is ultimately going to be overwritten when the build scripts normalize all of the mod manifests. I think this should be dropped.

`depends_on` key
---
See https://github.com/aclist/kbin-kes/discussions/495. This is probably no longer necessary if we are abstracting dependencies into their own files.

`catch_reset` key
---
This was originally designed for the `resize_text` mod, and allows the author to provide an array of numerical input fields that will be reset to their initial value when a reset button (one per mod) is pushed. The rationale was that this singular reset button could be used to restore all of the number fields to ther initial state as defined in the manifest. Now that `resize_text` uses a different logic, this entire functionality might be unnecessary. Or should it be kept around for some other use case?

Should this be expanded so that any mod can reset all of its own settings, not just number types? If so, why does the author have to provide an array of input fields ahead of time? Wouldn't it be better to set up the logic on the MES side, so that a reset button will always revert all of the mod's settings? If so, could this be changed so that the portion of the manifest asking whether you want to add a reset button is just a boolean option like `"reset": true`?

safeGM methods
---
safeGM currently uses the syntax `safeGM("method", "other params")`, where the first parameter is a stringwise rendering of the desired internal GM method. It seems like it would be more idiomatic and usable to make these actual methods under safeGM, such as `safeGM.addStyle("params")`.

Greasemonkey `setClipboard` method
---
There is no clear benefit to using this over the native `navigator.clipboard` property, and it is more complicated to use. Support for this should be dropped IMO.

Enum syntax
--- 
Enums are currently structured as, e.g., `Mbin.Microblog`. For better conformity and clarity that these are not methods, but enumerated constants, the syntax should be changed to `Mbin.MICROBLOG`. Similarly for subdomains such as `Mbin.Thread.Comments`, the final enumeration itself should be capitalized to `Mbin.Thread.COMMENTS`.

aclist:
[...]

PR #485, opened 2025-03-02
feat: debug bar

aclist:
Implements #459

Adds `helpers/debug.js`, which provides three public functions: 

- `debugBar()`: returns an HTMLElement which consists of a container with inner grid and a series of buttons/widgets arrayed in it. These buttons triggers various functionality, explained below.
- `isDebugBarEnabled()`: boolean test of whether the localStorage key `mes-debugbar["enabled"]` is true
- `toggleDebug()`: writes the state of the debug bar to localStorage and reloads the window, allowing the debug bar to be injected/removed and its settings applied/unapplied on pageload. This is done to show a "clean run" of what the page loads like when the debug bar's block filters are applied or unapplied.

The debug bar allows the user to block or unblock various mods from triggering, irrespective of the user's persistent settings in MES. This allows you to test/troubleshoot various mod interactions without destructively changing your usual settings. Debug filters are stored in a separate namespace.

If the debug bar is enabled, the underlying MES enable/disable state is ignored, and the debug bar's filters take precedence completely. This allows you to test mods you don't necessarily have enabled in your main MES settings.

>[!NOTE]
> Due to their complexity, the mod's granular settings saved in MES are still inherited when in debug mode. You can quickly access the mod's setting page via the debug bar to change these.

Pictured below, the debug bar is sandwiched below the main Mbin navbar and before the content area. An explanation of each widget follows.

debugbar

Mod selection area
---
- MES button: opens the MES modal to the page corresponding to the mod currently selected in the dropdown
- Version link: prints the MES version. Clicking this link opens a new tab to the GitHub issue page with the MES version number pre-filled
- Select menu: alphabetized list of all mods, using their internal snake case entrypoint names. Mods are grouped and numbered according to blocked/unblocked state, and the contents of this dropdown are updated whenever a mod's state changes. (In practice, the entire select menu is removed, repopulated, and reinserted).

Button area
---
- Block/unblock mod: toggles the state of the currently selected mod in the dropdown. This button updates dynamically depending on the underlying state of the dropdown's currently selected index.
- Block all: sets all mods to the blocked state
- Unblock all: sets all mods to the unblocked state
- Apply changes: after new values are saved to localStorage, triggers a page reload. Any change to the state of a mod or mods will highlight this button, indicating that there are unapplied changes
[...]

Issue #456, opened 2025-02-19
[CHORE] Add these paths to getPageType()


aclist:
### Browser

Firefox-based

### Script manager

Tampermonkey

### KES version

4.3.0-beta.35

### Describe the bug

```
https://<DOMAIN>/new
https://<DOMAIN>/new/article
https://<DOMAIN>/new/photo
https://<DOMAIN>/newMagazine
```

Mbin.New.Link
Mbin.New.Thread
Mbin.New.Photo
Mbin.New.Magazine

aclist:
Add other paths described in #419

PR #503, opened 2025-03-10
chore: add sphinx documentation

### LINKED ISSUE #317: [DOCS] Specify reserved variables that mods should not set
- [x] Explicitly list these values upfront in documentation
- [x] Add tests to scan mods for illegal values

NB: `settings["state"]` (JS) and `"key": "state"` (JSON) are reserved

### LINKED ISSUE #348: [DOCS] Expand troubleshooting section in docs with common issues
https://developer.chrome.com/docs/extensions/reference/api/userScripts#developer_mode_for_extension_users

### LINKED ISSUE #406: [DOCS] Document page enums introduced in #404
### Browser

Firefox-based

### Script manager

Tampermonkey

### KES version

4.3.0-beta.18

### Describe the bug

Add the page enums to the API section of the documentation.

### LINKED ISSUE #455: [DOCS] Document Log enums
### Browser

Firefox-based

### Script manager

Tampermonkey

### KES version

4.3.0-beta.33

### Describe the bug

Document enums introduced in #451

aclist:
This PR overhauls the documentation as discussed in https://github.com/aclist/kbin-kes/issues/454 and https://github.com/aclist/kbin-kes/discussions/362#discussioncomment-12397117

A README.md, Makefile, and other build scripts are provided.

Migration to Sphinx
---
Documentation was ported from Asciidoctor to Sphinx, although this was ultimately a near total-rewrite. There were many vague, missing or out of date sections in the old documentation, and the API reference was incomplete at best.

Rather than recycle assumptions from the old documentation, the new documentation has overall been written from the standpoint of the 5.0.0 release and forward-looking changes. Not all of those changes are in yet, so the documentation represents where MES is going, rather than where it is. As such, all outstanding framework changes must be merged **before** this documentation, including, for example, https://github.com/aclist/kbin-kes/pull/490.

Document new enums
---
- Resolves https://github.com/aclist/kbin-kes/issues/406
- Resolves #455
- Additionally, documents new `Triggers` enums introduced in https://github.com/aclist/kbin-kes/pull/497
- Also documents new `Theme` enums introduced in https://github.com/aclist/kbin-kes/pull/485/commits/3b2c37fbae3925efec1ba232d9922add3369b4d4 (pending merge)
- Also resolves https://github.com/aclist/kbin-kes/issues/317

Utility function reference
---
Functions are more richly documented now, with all params, return types, and examples listed. Cross-references (xrefs) are provided to jump to related function definitions or enumerations. A custom style has been applied to make function and method xrefs more clearly distinguishable from builtins (which are not links) on hover.

safeGM
---
safeGM has been given its own section in the documentation, rewritten from the vantage point that it is going to be refactored to support methods, as described in https://github.com/aclist/kbin-kes/issues/502.

Enums
---
Similarly to the above, the enums have been written from the vantage point that the uppercase notation from #502 is going to be used.

MES enums are not class-based. They are naive frozen objects with constant integer values. This datatype was not clearly supported in Sphinx out of the box, so I faced the choice to either extend it by writing a custom directive for the JS domain, or using the `.. js:attr::` directive to list the members as a series of attributes. Sphinx does support creating a global module under which attributes are namespaced, but this causes the attributes to not look and feel like traditional enum documentation ([PyGtk enums for reference](https://lazka.github.io/pgi-docs/Gtk-3.0/enums.html)).

To emulate this style, I opted for the workaround of appending a custom class to the enum namespace and members, respectively. CSS is then applied to give the members a discrete appearance and show the integer enumeration. Albeit a kludge, this is reasonably simple for our purposes and achieves the intended effect without additional work.

JSON schemas
---
JSON schemas have been introduced for the main manifest and for all subschemas describing settings menu types like `radio`, `select`, `color`, etc. Required and optional properties are explicitly listed now, and the datatype of each property is listed, as well as enumerations of which (and how many) values are permissible. This section also resolves https://github.com/aclist/kbin-kes/issues/317, stating upfront that the `state` key is reserved.

Rather than a loosely defined example JSON blurb, the schemas should make it absolutely clear how the JSON is to be formatted for every requested setting type, and links to the raw schema JSON are provided for reference.

Known issues
---
As a side effect of not namespacing the enums under module names (e.g., under default Sphinx behavior, the member `LOG = 1` would instead be listed as `Log.LOG = 1`, with the enumeration including the namespace), there is the possibility of collisions between `:js:attr` xrefs, since they share the same domain. For example, `Mbin.Thread.COMMENTS` might be seen as the same as `Mbin.Domain.COMMENTS`, depending on how the document is traversed during compilation. For the most part, Sphinx traverses auto xrefs correctly, but there was one xref that was getting resolved to the wrong internal anchor.

For now, I've added a monkey patch (`patch-1.sh`) in the directory root that simply seds the offending anchor and updates it to the correct ID to deduplicate it. There was only one xref collision.

The `fa-clipboard` icon shown on the main menu legend page does not match 1:1 with the clipboard icon in MES, I think because Mbin is using a different fontawesome version from the one being sourced in Sphinx. I'll try to see if this can be overridden, but it's a minor thing.

TODO
---
- [x] The majority of utility functions from `helpers/safeGM.js` were documented, but still to be added are the functions and methods around `loadMags()`from https://github.com/aclist/kbin-kes/pull/335
- [x] The page enums from #456 haven't been added yet.
- [x] There is ample documentation around the `Trigger` enums, but would like to add a tangible example of how this functionality works with an example code block
- [x] Add mention of the developer mode requirement from https://github.com/aclist/kbin-kes/issues/348


aclist:
@Pamasich to review this, I would recommend building the HTML files directly and inspecting the rendered page, as there are a lot of miscellaneous build artifacts and small files that won't provide much information on their own. If building locally proves to be an issue, I can host the docs on a subdirectory of GitHub pages somewhere.

aclist:
I added a couple of TODOs to this, so going to convert it back to a draft. If you want to pre-review it, I can incorporate any feedback on a rolling basis.

aclist:
> The fa-clipboard icon shown on the main menu legend page does not match 1:1 with the clipboard icon in MES, I think because Mbin is using a different fontawesome version from the one being sourced in Sphinx. I'll try to see if this can be overridden, but it's a minor thing.

The library https://github.com/sphinx-contrib/icon uses more modern (fa5 vs fa4) icons, but has an outstanding bug that prevents the compiled HTML from using relative paths.

aclist:
This is complete on testing, with the proviso that the public HTML must be compiled and uploaded to GitHub pages once ready.

PR #506, opened 2025-03-13
refactor: changed how loadMags works for mods which want the callback to only run once

Pamasich:
As announced in #468, this PR changes how loadMags deals with callbacks that should only be run once. Instead of loadMags taking a parameter which makes the callback only execute once, it now runs the callback twice all the time, but passes a new boolean parameter which indicates whether it's the final call (true) or not (false). This enables mods to have code for both scenarios — code which can run both times, and code which cannot.


aclist:
Great idea. I'll incorporate this into the softblock PR as well.

Can I approve this?

Pamasich:
Yeah, this is finished.

aclist:
I'm not sure why I kept stating that softblock needs to use this function. Softblock's implementation of loadMags() is a totally different function that fetches the list of mags the user softblocked, not the full list of the user's subscriptions. At no point does softblock need to fetch the full list of subscriptions; it's not related at all. If anything, softblock could implement a similar "wait for all mags to load" functionality, but it already has its own async methods and the user would have to block a lot of magazines for this to start being a problem, if at all. Besides, the loading functionality is solely confined to this mod. Furthermore, there is no need to fetch multiple pages of remote XML content, since the blocklist is contained in one place.

PR #507, opened 2025-03-13
fix: prevent adding CSS to document head with identical ID

aclist:
Resolves the issue discussed in #501 regarding multiple stylesheets stacking on top of each other.

Preferably, mods should prefix the mod name in front of each sheet ID when calling safeGM, e.g., `softblock-custom-css` rather than vague descriptors like `hide-table`.


Pamasich (approved):
Works perfectly now.

